PR #125, opened 2025-03-04
Front end for CWT credentials

w4ll3:
## Description

Mobile side of CWT changes

## Tested

Tested at Iphone 14+, Iphone 13, Samsung A54, with local deployment and remote


ios/Showcase/Targets/AppUIKit/Sources/verifier/VerifyCwt.swift
```diff
@@ -0,0 +1,58 @@
...
+                    scanningType: .qrcode,
+                    onCancel: {
+                        path.removeLast()
+                    },
+                    onRead: { code in
+                        Task {
+                            do {
+                                credentialPack = CredentialPack()
+                                let cwt = try Cwt.newFromBase10(payload: code)
+                                _ = credentialPack!.addCwt(cwt: cwt)
```
cobward, line 31: Why do we add it to a credentialPack? We're just verifying, not storing

w4ll3: It's for display only, we are indeed not storing.

Juliano1612 (approved):

PR #130, opened 2025-03-11
Add READMEs to iOS projects

cobward:
## Description

Add READMEs to the iOS projects with instructions for how to initialize the xcode projects.

### Other changes

N/A

### Optional section

N/A

## Tested

N/A


sbihel (commented):
If you think it's better to have instructions in each project, please remove the sections in CONTRIBUTING.md and add links to these READMEs

PR #137, opened 2025-04-04
use SSI revision patch for pre-release

Ryanmtate:
## Description

Use patched preview of ssi `0.10.3`. Use tagged release once available.

### Other changes

Updates anyhow dependency to match SSI.

### Other

Over 500 lines due to `Cargo.lock`

## Tested

Tested as a maven local release on android.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -151,6 +151,15 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


Juliano1612 (approved):
Tested with a local Sprucekit Showcase app version and the presentation is working again!

PR #140, opened 2025-04-15
Add presentation signing function

theosirian:
## Description

Adds a function to sign one or more `ParsedCredential`s into a json-ld `VerifiablePresentation`.

Breaking changes to:
- `DidMethod::did_from_jwk`: now returns `DIDBuf`, but `DidMethodUtils::did_from_jwk` is unchanged and still returns a `String`.
- `DidMethod::vm_from_jwk`: now returns `DIDVerificationMethod`, but `DidMethodUtils::vm_from_jwk` is unchanged and still returns a `String`.

## Tested

Tested through app functionality staying the same after replacing `DIDKit` signing by this new presentation signing.

### DIFF
--- a/rust/src/oid4vp/holder.rs
+++ b/rust/src/oid4vp/holder.rs
@@ -462,6 +462,7 @@ pub(crate) mod tests {
             DidMethod::Key
                 .vm_from_jwk(&self.jwk())
                 .await
+                // SAFETY: The JWK should always be well-formed and this method should not panic.
                 .unwrap()
                 .id
                 .to_string()
--- a/rust/src/oid4vp/holder.rs
+++ b/rust/src/oid4vp/holder.rs
@@ -459,11 +459,19 @@ pub(crate) mod tests {
         }
 
         async fn verification_method(&self) -> String {
-            DidMethod::Key.vm_from_jwk(&self.jwk()).await.unwrap()
+            DidMethod::Key
+                .vm_from_jwk(&self.jwk())
+                .await
+                .unwrap()
+                .id
+                .to_string()
         }
 
         fn did(&self) -> String {
-            DidMethod::Key.did_from_jwk(&self.jwk()).unwrap()
+            DidMethod::Key
+                .did_from_jwk(&self.jwk())
+                .unwrap()
+                .to_string()
         }
 
         fn cryptosuite(&self) -> CryptosuiteString {
--- a/rust/src/oid4vp/holder.rs
+++ b/rust/src/oid4vp/holder.rs
@@ -471,6 +471,7 @@ pub(crate) mod tests {
         fn did(&self) -> String {
             DidMethod::Key
                 .did_from_jwk(&self.jwk())
+                // SAFETY: The JWK should always be well-formed and this method should not panic.
                 .unwrap()
                 .to_string()
         }
--- a/rust/src/presentation/mod.rs
+++ b/rust/src/presentation/mod.rs
@@ -77,18 +77,24 @@ impl JsonLdPresentationBuilder {
             credentials
                 .into_iter()
                 .map(|c| match &c.inner {
-                    ParsedCredentialInner::MsoMdoc(_) => unimplemented!(),
-                    ParsedCredentialInner::JwtVcJson(jwt_vc_json) => {
-                        serde_json::Value::String(jwt_vc_json.jws.clone().into_string())
+                    ParsedCredentialInner::MsoMdoc(_) => {
+                        Err(PresentationBuilderError::UnsupportedCredentialFormat)
                     }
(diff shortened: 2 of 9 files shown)


rust/src/oid4vp/holder.rs
```diff
@@ -459,11 +459,19 @@ pub(crate) mod tests {
...
+                .id
+                .to_string()
         }
 
         fn did(&self) -> String {
-            DidMethod::Key.did_from_jwk(&self.jwk()).unwrap()
+            DidMethod::Key
+                .did_from_jwk(&self.jwk())
+                .unwrap()
+                .to_string()
```
Ryanmtate, line 474: I realize we were calling `.unwrap()` here. We should probably handle this error more gracefully.

Ryanmtate (approved):
Nice work, @theosirian ! Just a minor nit on the `.unwrap()` calls.

rust/src/oid4vp/holder.rs
```diff
@@ -459,11 +459,19 @@ pub(crate) mod tests {
         }
 
         async fn verification_method(&self) -> String {
-            DidMethod::Key.vm_from_jwk(&self.jwk()).await.unwrap()
+            DidMethod::Key
+                .vm_from_jwk(&self.jwk())
+                .await
+                .unwrap()
```
Ryanmtate, line 465:
```suggestion
                // SAFETY: The JWK should always be well-formed and this method should not panic.
                .unwrap()
```

rust/src/oid4vp/holder.rs
```diff
@@ -459,11 +459,19 @@ pub(crate) mod tests {
...
+                .id
+                .to_string()
         }
 
         fn did(&self) -> String {
-            DidMethod::Key.did_from_jwk(&self.jwk()).unwrap()
+            DidMethod::Key
+                .did_from_jwk(&self.jwk())
+                .unwrap()
```
Ryanmtate, line 473:
```suggestion
                // SAFETY: The JWK should always be well-formed and this method should not panic.
                .unwrap()
```

rust/src/oid4vp/holder.rs
Ryanmtate: Tiago: Might need an alternative version that takes an actual `JWK` object instead of the `&str`. Also can improve safety using Results and strict types here. Save for a later PR (out of scope for this PR).

rust/src/presentation/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+                    eip712_loader: (),
+                },
+                vp,
+                &resolver,
+                self,
+                params,
+                Default::default(),
+            )
+            .await
+            .unwrap();
```
cobward, line 124: Return an error instead of unwrap

rust/src/presentation/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+        );
+
+        let mut params = ProofOptions::from_method(IriBuf::new(vm)?.into());
+
+        params.proof_purpose = self.proof_purpose;
+        params.challenge = self.challenge.to_owned();
+        params.domains = self.domain.to_owned().map(|d| vec![d]).unwrap_or_default();
+
+        let resolver = VerificationMethodDIDResolver::new(AnyDidMethod::default());
+        let suite = AnySuite::pick(&key, params.verification_method.as_ref()).unwrap();
```
cobward, line 101: And here

rust/src/presentation/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+                    ParsedCredentialInner::JwtVcJson(jwt_vc_json) => {
+                        serde_json::Value::String(jwt_vc_json.jws.clone().into_string())
+                    }
+                    ParsedCredentialInner::JwtVcJsonLd(jwt_vc_json_ld) => {
+                        serde_json::Value::String(jwt_vc_json_ld.jws.clone().into_string())
+                    }
+                    ParsedCredentialInner::VCDM2SdJwt(_) => unimplemented!(),
+                    ParsedCredentialInner::LdpVc(ldp_vc) => ldp_vc.raw.clone(),
+                    ParsedCredentialInner::Cwt(_) => unimplemented!(),
+                })
```
cobward, line 90: return errors instead of unimplemented!(), which panics

rust/src/presentation/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+            proof_purpose: proof_purpose.unwrap(),
+            challenge,
+            domain,
+            signer: Arc::new(signer),
+            context_map,
+        }
+        .into()
+    }
+
+    pub async fn issue_presentation(
```
cobward, line 64: Might make more sense to just implement this only for the credential types  that can be ld presentations

cobward (changes requested):

cobward (approved):

PR #143, opened 2025-04-30
Implement Digital Credentials API support for Android

cobward:
## Description

Implement support for DC-API to the android code:
* Support DCQL over OpenID4VP (draft 24) over DC-API in the rust code.
* Add DC-API integrations in the android library code.
* Integrate in the Showcase app.

### Other changes

Fixed the constant rebuilding of the rust code in the android priject, now only does so on updates to the rust code.

### Optional section

* TODO: Use on-main commit of openid4vp after the PR is merged.
* If this PR exceeds 500 lines, please explain why:
  * The rust, and kotlin pieces can be split up and merged independently, but ultimately this is new functionality that requires all of the code to function.

## Tested

You can test using one of the interop event verifiers, such as https://funke.animo.id


rust/src/credential/mdoc.rs
```diff
@@ -223,3 +224,54 @@ pub enum MdocEncodingError {
...
+            Some(serde_json::Value::Number(num))
+        }
+        ciborium::Value::Bool(b) => Some(serde_json::Value::Bool(*b)),
+        ciborium::Value::Null => Some(serde_json::Value::Null),
+        ciborium::Value::Tag(_, value) => to_json_for_display(value),
+        _ => {
+            tracing::warn!("unsupported value type: {:?}", value);
+            None
+        }
+    }
```
Ryanmtate, line 276: Worth moving these to `isomdl` crate?

rust/src/oid4vp/dc_api/prepare_response.rs
```diff
@@ -0,0 +1,39 @@
...
+    pub fn new(origin: String, client_id: String, nonce: String) -> Result<Self> {
+        let handover_info = HandoverInfo(origin, client_id, nonce);
+        let handover_info_bytes = cbor::to_vec(&handover_info)?;
+        let handover_info_hash = ByteStr::from(Sha256::digest(handover_info_bytes).to_vec());
+        Ok(Handover(
+            "OpenID4VPDCAPIHandover".to_string(),
+            handover_info_hash,
+        ))
+    }
+}
```
Ryanmtate, line 27: Does it make sense to combine these with the `Handover::OID4VP` type in `isomdl` crate?

Ryanmtate (approved):
Nice work, @cobward ! Some minor comments about whether it makes sense to add some of these methods to the `isomdl` crate.

android/MobileSdk/build.gradle.kts
```diff
@@ -145,4 +147,18 @@ dependencies {
     testImplementation("org.json:json:20230618")
     androidTestImplementation("com.android.support.test:runner:1.0.2")
     androidTestImplementation("com.android.support.test.espresso:espresso-core:3.0.2")
+
+    // DC-API dependencies
+    implementation("androidx.credentials:credentials:1.5.0")
+    implementation("androidx.credentials.registry:registry-digitalcredentials-preview:1.0.0-alpha01")
+    implementation("androidx.credentials.registry:registry-digitalcredentials-mdoc:1.0.0-alpha01")
+    implementation("androidx.credentials.registry:registry-provider:1.0.0-alpha01")
+    implementation("androidx.credentials.registry:registry-provider-play-services:1.0.0-alpha01")
```
Juliano1612, line 156: Not sure if this makes sense here, but looks like the version should always be the same

```suggestion
    val androixCredentialsRegistryVersion = "1.0.0-alpha01"
    implementation("androidx.credentials.registry:registry-digitalcredentials-preview:$androixCredentialsRegistryVersion")
    implementation("androidx.credentials.registry:registry-digitalcredentials-mdoc:$androixCredentialsRegistryVersion")
    implementation("androidx.credentials.registry:registry-provider:$androixCredentialsRegistryVersion")
    implementation("androidx.credentials.registry:registry-provider-play-services:$androixCredentialsRegistryVersion")
```

android/Showcase/src/main/java/com/spruceid/mobilesdkexample/wallet/HandleMdocOID4VPView.kt
```diff
@@ -243,9 +245,10 @@ fun MdocFieldSelector(
             textAlign = TextAlign.Center
         )
 
+        val columnMod = if (fillHeight) Modifier.fillMaxSize() else Modifier
```
Juliano1612, line 248: This is not usual. It's better to accept the `modifier: Modifier = Modifier` as parameter and use this if necessary. In this case, you will only need to change the MdocFieldSelector in the `HandleMdocOID4VPView.kt`to
```
MdocOID4VPState.SelectiveDisclosure ->
            MdocFieldSelector(
                match = selectedMatch!!,
                onContinue = {  ... },
                modifier = Modifier.fillMaxSize(),
                onCancel = { onBack() }
            )
```

android/Showcase/src/main/java/com/spruceid/mobilesdkexample/MainActivity.kt
```diff
@@ -182,4 +183,17 @@ class MainApplication : Application() {
...
+//    lateinit var dcApiRegistry: DcApiRegistry
+//
+//    override fun onCreate() {
+//        super.onCreate()
+//        try {
+//            dcApiRegistry = DcApiRegistry(this)
+//        } catch (e: Exception) {
+//            Log.e("MainApplication", "failed to setup DcApiRegistry: $e")
+//        }
+//    }
```
Juliano1612, line 198:
```suggestion
```

android/Showcase/build.gradle.kts
```diff
@@ -88,8 +89,22 @@ dependencies {
...
     debugImplementation("androidx.compose.ui:ui-tooling")
     debugImplementation("androidx.compose.ui:ui-test-manifest")
+
+    // DC-API dependencies
+    implementation("androidx.credentials:credentials:1.5.0")
+    implementation("androidx.credentials.registry:registry-digitalcredentials-preview:1.0.0-alpha01")
+    implementation("androidx.credentials.registry:registry-digitalcredentials-mdoc:1.0.0-alpha01")
+    implementation("androidx.credentials.registry:registry-provider:1.0.0-alpha01")
+    implementation("androidx.credentials.registry:registry-provider-play-services:1.0.0-alpha01")
+    implementation("com.google.android.gms:play-services-identity-credentials:16.0.0-alpha06")
```
Juliano1612, line 103: The same as I said before.

Juliano1612 (changes requested):
It looks good, but I added some suggestions. 

Also, the "Generate mDL" feature is not working on my side. Can you confirm if it's working for you?

rust/src/oid4vp/dc_api/prepare_response.rs
cobward: I don't think so, probably we should just remove the `Handover::OID4VP` variant from `isomdl`. We possibly need to make some decisions about what the responsibilities of the different libraries are, but IMO `isomdl` is currently:
* Data structures according to 18013-5.
* 18013-5 presentation transaction logic.

The OpenID4VP stuff is constantly changing, so for now I would prefer to keep it separate.

rust/src/credential/mdoc.rs
cobward: I'll keep it here for now, it's pretty specific to the wallet unless we think it's useful in verifier code?

sbihel (approved):
Did not review the code but all my tests were successful.

PR #145, opened 2025-05-08
Add Licenses

cobward:


sbihel (approved):

sbihel (commented):
Actually, it seems like not all packages have the licenses specified, could you add it to Package.swift as well? The apps don't have licenses either, but I'm not sure it possible to embed licenses in an iOS or Android app

wyc (approved):

PR #149, opened 2025-05-20
Toggle for production and staging HACI in showcase app

Juliano1612:
## Description

This allows enable and disable dev mode when using the Showcase app. The only feature impacted at this moment is the HACI, but the implementation allows extending this for other features in the future. The dev mode configuration is not persistent, so when you (re)launch the app, it will always be set to production.

## Tested

To enable/disable dev mode you can go to the wallet settings and click the Enable/Disable Dev Mode button.


w4ll3 (approved):

w4ll3:
Tested on Android A54, working just fine E2E with verifier (staging, new certs are not in prod yet)

PR #154, opened 2025-06-09
Lazy-load available endpoints from Issuance and Wallet services

SeaSkyThe:
## Description

This PR introduces lazy fetching and caching of available endpoints from the Issuance and Wallet services from [Grove](https://github.com/spruceid/grove).

### Other changes

N/A

### Optional section

* Reviewers, please pay special attention to...
* If this PR exceeds 500 lines, please explain why

## Tested

Not yet

## TODO

- [X] Issuance service client endpoints lazy-load
- [ ] Wallet service client endpoints lazy-load
- [ ] Wallet service base-url lazy-load (get from issuance service /.well-known)
- [ ] Test both app versions

### DIFF
--- a/rust/src/haci/issuance_service_client.rs
+++ b/rust/src/haci/issuance_service_client.rs
@@ -1,3 +1,6 @@
+use std::collections::HashMap;
+use std::sync::RwLock;
+
 use crate::haci::http_client::HaciHttpClient;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
@@ -24,6 +27,10 @@ pub enum IssuanceServiceError {
     /// Internal error
     #[error("Internal error: {0}")]
     InternalError(String),
+
+    /// Missing endpoint
+    #[error("Endpoint key does not exists: {0}. Available keys: {1}")]
+    MissingEndpoint(String, String),
 }
 
 #[derive(Debug, Serialize, Deserialize)]
@@ -41,6 +48,7 @@ pub struct CheckStatusResponse {
 pub struct IssuanceServiceClient {
     client: HaciHttpClient,
     base_url: String,
+    endpoints: RwLock<Option<HashMap<String, String>>>,
 }
 
 #[uniffi::export(async_runtime = "tokio")]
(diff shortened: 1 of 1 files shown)


cobward:
> I have three concerns:
> 
>     1. About supporting redirect (https://proofing.haci.spruceid.xyz/ -> https://proofing.grove.spruceid.xyz/), isn't easier for us to do that at the DNS/Infra level?
> 
>     2. How do I fix the iOS step on the CI?
> 
>     3. Should I make the IssuanceClientService and WalletClientService backwards compatible so we don't necessarily need to update Grove and then the App in a specific order?

1. Yep exactly, we will do it there. As long as this will work after we setup the redirect then I think there's no problems.

2. Juliano answered that one :)

3. App updates are more work, so I think it's fine to just update Grove first, and then release this change. Otherwise we would have to come back and clean it up later.

rust/src/haci/issuance_service_client.rs
```diff
@@ -34,13 +41,15 @@ struct NewIssuanceResponse {
...
-    openid_credential_offer: String,
+    openid_credential_offer: Option<String>,
+    proofing_url: Option<String>,
 }
 
 #[derive(uniffi::Object)]
 pub struct IssuanceServiceClient {
     client: HaciHttpClient,
     base_url: String,
+    endpoints: RwLock<Option<HashMap<String, String>>>,
```
cobward, line 52: Locks are not ideal, particularly across FFI in the case that they are poisoned. There's a few places in the code that can panic due to the use of RwLocks here.

I think a `tokio::sync::OnceCell` would be a better option here:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=39e09f0dd0f2e28157b2cc77593ad38b

cobward (changes requested):

cobward (approved):
Approved, but lets hold off merging this until we have deployed the backend changes to production

SeaSkyThe:
I will be closing this PR, the followup is: #162

PR #162, opened 2025-06-30
Lazy-load available endpoints from Issuance and Wallet services

SeaSkyThe:
## Description

Rebased version of #154. 

### Other changes

Same as original PR

### Optional section

Same as original PR

## Tested

Same as original PR

### DIFF
--- a/rust/src/haci/issuance_service_client.rs
+++ b/rust/src/haci/issuance_service_client.rs
@@ -75,7 +75,11 @@ impl IssuanceServiceClient {
     /// * `base_url` - The base URL of the issuance service
     #[uniffi::constructor]
     pub fn new(base_url: String) -> Self {
-        let actual_url = base_url.trim().strip_suffix('/').unwrap_or(&base_url).to_string();
+        let actual_url = base_url
+            .trim()
+            .strip_suffix('/')
+            .unwrap_or(&base_url)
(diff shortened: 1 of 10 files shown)


rust/src/haci/issuance_service_client.rs
```diff
@@ -215,18 +349,94 @@ mod tests {
...
+            .expect(1)
+            .mount(&mock_server)
+            .await;
+
+        // Mock successful status check response
+        Mock::given(method("GET"))
+            .and(path(format!("/issuance/{issuance_id}/status")))
+            .respond_with(ResponseTemplate::new(200).set_body_json(json!({
+                "state": "ProofingRequired",
+                "proofing_url": "http://www.teste.com.br/"
```
Juliano1612, line 381:
```suggestion
                "proofing_url": "http://www.test.com/"
```

Juliano1612 (approved):
LGTM! just added a suggestion to update the test url

rust/src/haci/issuance_service_client.rs
SeaSkyThe: My brazilian side woke up 😅. 

Since this is a test, I'll leave it as is for now so we don't delay things. I can update it in a future PR.

PR #165, opened 2025-07-08
Envelope encryption/decryption credentials

Juliano1612:
## Description

This implements hybrid en/decryption methods to improve performance for big credentials.

Loading two credentials of 5MB each was taking 13+ seconds due to Android Keystore's poor performance with large payloads. Hardware Security Modules (HSM/TEE/Strongbox) have significant overhead for bulk data encryption, causing unacceptable load times on the initial screen.

This implemented hybrid encryption (envelope pattern) where Android Keystore protects only a 32-byte AES key while large credential data is encrypted with software AES/GCM. This follows industry standards used by Google Cloud KMS and AWS KMS. The solution maintains backward compatibility with existing encrypted data through automatic format detection.

### Other changes

N/A

### Optional section

N/A

## Tested

It's possible to test this by issuing a [VC Playground](https://vcplayground.org/issuer) Driver's License and checking how long it takes to save and load the credential. Unfortunately, it's broken, but I can help with some setup to make it locally.

### DIFF
--- a/android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/CredentialPack.kt
+++ b/android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/CredentialPack.kt
@@ -109,7 +109,7 @@ class CredentialPack {
     /**
      * Try to add a credential in any supported format (standard credential or mdoc).
      * Attempts to parse as standard credential first, then as mdoc if that fails.
-     * 
+     *
      * @param rawCredential The raw credential data as a string
      * @param mdocKeyAlias The key alias to use if parsing as mdoc is needed
      * @return List of parsed credentials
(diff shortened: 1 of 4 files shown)


cobward (commented):
How does this affect applications that already use the StorageManager? Can we do a migration?

Juliano1612:
> How does this affect applications that already use the StorageManager? Can we do a migration?

All new credentials will be encrypted using the hybrid approach. But when decrypting, I added a check to choose the method https://github.com/spruceid/sprucekit-mobile/pull/165/files#diff-0900499bd775fd9651934a101404917b0a6859a5dd7b3a02ebff3909082f61d3R45

So it is backward compatible.

cobward (approved):

PR #177, opened 2025-07-16
Dismiss Toast on tap

Juliano1612:
## Description

This updates the toast component to be dismissed when the user taps it.

### Other changes

This also adds a feature to copy the toast message to the clipboard if the user long-presses it.

### Optional section

N/A

## Tested

You can click to generate a mock mDL and interact with the toast.


SeaSkyThe (approved):
Tested both in Android an iOS. LGTM!

PR #182, opened 2025-07-21
Universal scanner UI

Juliano1612:
## Description

This updates the universal scanner UI according to the latest Figma designs.

### Other changes

N/A

### Optional section

N/A

## Tested

You can check the SpruceKit Showcase QR code scanners


SeaSkyThe (approved):
Tested on both Android and iOS and it looks good.

PR #202, opened 2025-09-18
Feat/cose sign1 signing key method

Ryanmtate:
## Description

This pull request introduces a new utility function for COSE_Sign1 signing in the Rust SDK, and increments the crate version. The main change is the addition of a `cose_sign1` function, which creates a COSE_Sign1 object by CBOR-encoding a payload, wrapping it with a Tag24, and signing it with a provided signing key. This will help facilitate secure payload signing in client applications.

**New COSE_Sign1 signing utility:**

* Added a new `cose_sign1` function to `CryptoCurveUtils`, which takes a signing key and a payload, wraps the payload in CBOR Tag24, constructs a COSE_Sign1 object, signs it, and returns the signed CBOR-encoded bytes.
* Imported necessary dependencies from `isomdl` and `ssi` crates to support COSE_Sign1 construction and CBOR encoding.

### Other changes

* Adds a cose key encoding for ec2 p256 curve to key manager

**Version bump:**

* Bumped crate version in `Cargo.toml` from `0.12.9` to `0.12.10` to reflect the new functionality.

### Optional section

Note, I might add encryption methods as well.

## Tested

Tested in myco app

### DIFF
--- a/ios/MobileSdk/Sources/MobileSdk/KeyManager.swift
+++ b/ios/MobileSdk/Sources/MobileSdk/KeyManager.swift
@@ -3,7 +3,8 @@ import Foundation
 import Security
 import SpruceIDMobileSdkRs
 
-public class KeyManager: NSObject, SpruceIDMobileSdkRs.KeyStore, ObservableObject {
+public class KeyManager: NSObject, SpruceIDMobileSdkRs.KeyStore, ObservableObject, @unchecked
+    Sendable {
     /// Migrate keys between access groups. For more information see
     /// https://developer.apple.com/documentation/Security/kSecAttrAccessGroup
     public func migrateToAccessGroup(oldAccessGroup: String, newAccessGroup: String) throws {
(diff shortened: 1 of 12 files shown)


ios/MobileSdk/Sources/MobileSdk/KeyManager.swift
```diff
@@ -146,6 +146,29 @@ public class KeyManager: NSObject, SpruceIDMobileSdkRs.KeyStore, ObservableObjec
...
+    
+    /**
+     * Returns the public key as a CBOR-encoded COSE key byte array
+     */
+    public static func coseKeyEc2P256PubKey(id: String) -> [UInt8]? {
+        guard let key = getSecretKey(id: id) else { return nil }
+
+        guard let publicKey = SecKeyCopyPublicKey(key) else {
+            return nil
+        }
```
jessmelanson, line 158: Small/unnecessary change, but you could combine the two guard statements with a comma so you only have to call `return nil` once.

jessmelanson (approved):
Looks good. Some weird indentation/formatting changes in the Kotlin code, maybe the linter will pick up on them.

Ryanmtate:
Ideally, this PR wouldn't include so many formatting changes. We may want to consider using something like [`ktlint`](https://github.com/pinterest/ktlint) to standardize the formatting.

PR #205, opened 2025-10-01
[ANDROID] Fix L2Cap

SeaSkyThe:
## Description

- Accept even PSM values with warning (some BLE stacks incorrectly allocate even PSMs despite spec requiring odd)
- Use dedicated Thread instances for L2CAP I/O instead of coroutines to prevent thread pool exhaustion
- Add exit signal in disconnect() to properly terminate L2CAP write threads and prevent resource leaks
- Remove redundant 0x02 termination when peer closes L2CAP connection to prevent cross-session contamination
- Fix GATT fallback to use reportLog() instead of reportError() to avoid premature session termination
- Fix disconnect order to call disconnect() before close() per Android BLE best practices

### Other changes

N/A

### Optional section

Since I’m implementing the spec version that uses a header to indicate the data size via L2CAP, which is a new part of the spec, I’ll disable it after the review. We can enable it later once more parties have implemented it.

## Tested

Tested both with L2Cap enabled and disabled, using the Multipaz app on Iphone. Both using SKS as Holder (Central and Peripheral) and Reader (Central and Peripheral)

Also tested with L2Cap disabled with GET Verify and mID Verify.

## TODO

- [ ] Disable L2Cap before merging

### DIFF
--- a/android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/ble/BleConfiguration.kt
+++ b/android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/ble/BleConfiguration.kt
@@ -20,7 +20,7 @@ data class BleConfiguration(
     val defaultMtu: Int = 23,  // BLE minimum
 
     // L2CAP configuration
-    val useL2CAP: L2CAPMode = L2CAPMode.IF_AVAILABLE,
+    val useL2CAP: L2CAPMode = L2CAPMode.NEVER,
     val l2capBufferSize: Int = 65536,  // 64KB
     val l2capReadTimeoutMs: Long = 500L,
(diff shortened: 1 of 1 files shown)


android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/ble/GattServer.kt
```diff
@@ -46,7 +45,7 @@ class GattServer(
     private var isReaderServer: Boolean = false,
     private val config: BleConfiguration = BleConfiguration()
 ) {
-
+    private val bleConfiguration = BleConfiguration()
```
w4ll3: This is already available via the `config` on the parameters.
```suggestion
```

w4ll3 (approved):

PR #207, opened 2025-10-08
[ACC 806] Style credential cards

SeaSkyThe:
## Description

Style credential cards to match SEDI designs

https://github.com/user-attachments/assets/ba540d6e-4b85-4880-9756-380f06e39828

### Other changes

N/A

### Optional section

- Readability is a little off, but I'm not sure what should I do to improve that

## Tested

- Build it and add different credentials to the wallet


SeaSkyThe:
> LGTM! In the future, we could have at least one more credential_bg, so we can use one only for mDLs and the other for regular VCs. Can you add a task to discuss it?

https://linear.app/spruceid/issue/ACC-819/different-backgrounds-to-different-credential-types

PR #230, opened 2025-11-10
separate the cbor encoding utils from the cose_sign1 method

Ryanmtate:
## Description

This pull request refactors how payloads are handled when signing with COSE_Sign1 and introduces a new utility for CBOR encoding. The main change is to decouple CBOR encoding from the signing process, making it the caller's responsibility to prepare the payload in the desired format before signing. Additionally, a new function is added to handle CBOR encoding and optional Tag24 tagging.

**Refactor of COSE_Sign1 signing flow:**

* The `cose_sign1` function no longer accepts an `is_cbor_payload` parameter and expects the payload to already be CBOR-encoded; the function's documentation and implementation are updated to reflect this change. [[1]](diffhunk://#diff-f567cff6baa49669b57e4cebbd048a5462e3bcdd80144f85cd99566fe4b755ddL81-L105) [[2]](diffhunk://#diff-f567cff6baa49669b57e4cebbd048a5462e3bcdd80144f85cd99566fe4b755ddL130-R116)

**New CBOR encoding utility:**

* Added a new `encode_to_cbor_bytes` function that encodes raw bytes as CBOR, with an option to tag the payload as Tag24. This centralizes encoding logic and makes payload preparation explicit for callers.

### Other changes

bumps minor version to 0.12.12.

### DIFF
--- a/rust/MobileSdkRs/Sources/MobileSdkRs/mobile_sdk_rs.swift
+++ b/rust/MobileSdkRs/Sources/MobileSdkRs/mobile_sdk_rs.swift
@@ -21935,7 +21935,7 @@ public func coseKeyEc2P256PublicKey(x: Data, y: Data, kid: Data)throws  -> Data 
 })
 }
 /**
- * This method accepts raw bytes to signed and included in a
+ * This method accepts raw bytes to be signed and included in a
  * COSE_Sign1 message.
  *
  * NOTE: The payload must be encoded to the desired format (e.g., CBOR bytes) BEFORE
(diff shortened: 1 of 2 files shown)


rust/MobileSdkRs/Sources/MobileSdkRs/mobile_sdk_rs.swift
```diff
@@ -21935,22 +21935,18 @@ public func coseKeyEc2P256PublicKey(x: Data, y: Data, kid: Data)throws  -> Data
 })
 }
 /**
- * This method encodes raw bytes as CBOR, tagging the payload as
- * a Tag24 data item and constructing a COSE_Sign1 object that is
- * signed by the signing key, with the signature included in the
- * CBOR bytes encoded COSE_Sign1 object returned.
+ * This method accepts raw bytes to signed and included in a
```
jessmelanson: Typo? "raw bytes to be signed" instead?

rust/src/crypto.rs
```diff
@@ -78,31 +78,17 @@ impl CryptoCurveUtils {
     }
 }
 
-/// This method encodes raw bytes as CBOR, tagging the payload as
-/// a Tag24 data item and constructing a COSE_Sign1 object that is
-/// signed by the signing key, with the signature included in the
-/// CBOR bytes encoded COSE_Sign1 object returned.
+/// This method accepts raw bytes to signed and included in a
```
Ryanmtate: NOTE: Will need to run the uniffi bindings again to ensure the comments are correctly added to the swift bindings.

```suggestion
/// This method accepts raw bytes to be signed and included in a
```

PR #234, opened 2025-11-17
CWT: Add status list check

SeaSkyThe:
## Description

Add CWT credential status checking using Status Lists.

**Added `status()` method to `Cwt`:**
  - Parameters:
    - `status_claim_key` - CBOR claim key containing status info (e.g., "65535")
    - `status_field_name` - Field name within the claim (e.g., "status_list")
  - Returns status code as `i16`:
    - `0` = VALID (credential is valid)
    - `1` = INVALID (credential is revoked)
    - `2` = SUSPENDED (credential is suspended)
    - `-1` = No status information available
  - Extracts `idx` and `uri` from CBOR structure
  - Fetches and decodes compressed status list using `JsonStatusList`
  
  For that to fully work with `SprucePass` we need to merge: https://github.com/spruceid/grove/pull/134 (but we can merge it without any problems)
 ### Usage
   
 ```kotlin
  val status = cwt.status()
  when (status.toInt()) {
      0 -> res[credentialId] = CredentialStatusList.VALID
      1 -> res[credentialId] = CredentialStatusList.INVALID
      2 -> res[credentialId] = CredentialStatusList.REVOKED
      else -> res[credentialId] = CredentialStatusList.UNDEFINED
  }
  ```
### Other changes

**New error variants:**
  - `StatusListFetch` - Network/fetch errors
  - `StatusListParse` - JSON parsing errors
  - `StatusListDecode` - Bit string decompression errors
  - `StatusIndexOutOfBounds` - Index exceeds status list size

### Optional section

When SSI adds native CWT status list support, this can be simplified to use the library's CWT-specific APIs instead of manual CBOR extraction

## Tested

iOS:

https://github.com/user-attachments/assets/98407647-47d6-47b4-9c00-de922411264e

Android:

https://github.com/user-attachments/assets/7b491d41-04fc-49c1-972e-005c2b47e2fe

## Todo:
- [X] Android
- [x] iOS

### DIFF
--- a/android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/CredentialPack.kt
+++ b/android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/CredentialPack.kt
@@ -220,9 +220,8 @@ class CredentialPack {
                 }
             }
             credential.asCwt()?.let {
-                Log.d("CredentialPack", "This is a cwt: ${it.credentialClaims()}")
                 try {
-                    val status = it.status(statusClaimKey = "65535", statusFieldName = "status_list")
+                    val status = it.status()
                     when (status.toInt()) {
                         0 -> res[credentialId] = CredentialStatusList.VALID
                         1 -> res[credentialId] = CredentialStatusList.INVALID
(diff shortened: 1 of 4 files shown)


android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/CredentialPack.kt
```diff
@@ -217,6 +219,20 @@ class CredentialPack {
                     res[credentialId] = CredentialStatusList.UNKNOWN
                 }
             }
+            credential.asCwt()?.let {
+                Log.d("CredentialPack", "This is a cwt: ${it.credentialClaims()}")
```
Juliano1612:
```suggestion
```

PR #249, opened 2025-12-16
[Flutter] Scanner

Juliano1612:
## Description

This PR introduces the Scanner API in the Flutter plugin that wraps the native iOS and Android SpruceKit Mobile SDK functionality.

### Other changes

N/A

### Optional section

N/A

## Tested

- Tested the Scanner


SeaSkyThe (approved):
For me it looks good! 

Just a concern, the scanner initialization can fail, right? What do you think about adding a `onError` callback to it? So we can deal with errors in the Flutter side? Do you think that's needed?

I'm also ok with leaving this PR open for now, and we can merge it after holidays if you feel like it.

PR #250, opened 2025-12-18
[Flutter] mDL Presentation 18013-5

Juliano1612:
## Description

This PR introduces the mDL Presentation and Spruce Util APIs in the Flutter plugin that wraps the native iOS and Android SpruceKit Mobile SDK functionality.

### Other changes

Also added the mDL presentation flow to the Flutter example app

### Optional section

N/A

## Tested

- Tested the presentation with SpruceKit Showcase or any other mDL reader

### DIFF
--- a/flutter/pigeons/spruce_utils.dart
+++ b/flutter/pigeons/spruce_utils.dart
@@ -15,7 +15,6 @@ import 'package:pigeon/pigeon.dart';
     dartPackageName: 'sprucekit_mobile',
   ),
 )
-
 /// Result of generating a mock mDL
 sealed class GenerateMockMdlResult {}
 


SeaSkyThe (commented):
I tested the setup and almost everything works as expected. The only issue I encountered is on iOS when presenting to the Android Multipaz test app. In this scenario, the request items are empty and the flow fails.

The same behavior occurs when presenting from the SpruceKit Showcase instead of the test app. This suggests that the issue is likely either in our iOS native implementation or in the version of the Multipaz Test app I’m using.

Other than that, all other presentations works fine (mID and sprucekit showcase). I think this is a good PR for @w4ll3  to take a look and test, if we don't have a problem waiting.

PR #257, opened 2025-12-19
take limit on filtered and sorted activity log entries

Ryanmtate:
## Description

This pull request refactors how activity log filtering and limiting are handled in the Rust backend. The main change is moving the enforcement of the `max_items` limit from the filtering logic into the main result processing pipeline, which simplifies the filter method and ensures consistent ordering before truncation.

**Refactoring of filtering and limiting logic:**

* The `filter` method in `ActivityLogFilterOptions` now only receives an `ActivityLogEntry` (not a tuple with an index) and no longer checks the `max_items` limit itself. [[1]](diffhunk://#diff-58285f52958a792ffeee62a7c7f3f7858524d96329332c3ea9cfae5ed109602aL70-R70) [[2]](diffhunk://#diff-58285f52958a792ffeee62a7c7f3f7858524d96329332c3ea9cfae5ed109602aL106-L113)
* The enforcement of the `max_items` limit is moved to the main result pipeline, applying `.take(max_items)` after sorting entries by date, ensuring the most recent entries are always included. This change is applied in both the cache-based and async-fetch code paths in `ActivityLog`. [[1]](diffhunk://#diff-58285f52958a792ffeee62a7c7f3f7858524d96329332c3ea9cfae5ed109602aL495-R493) [[2]](diffhunk://#diff-58285f52958a792ffeee62a7c7f3f7858524d96329332c3ea9cfae5ed109602aL629-R633)

- bumped version to 0.13.13

## Tested

TBT

### DIFF
--- a/rust/src/credential/activity_log.rs
+++ b/rust/src/credential/activity_log.rs
@@ -96,7 +96,7 @@ impl ActivityLogFilterOptions {
         if self
             .interacted_with
             .as_ref()
-            .is_some_and(|actor| *actor != entry.interaction_with)
+            .is_some_and(|actor| entry.interaction_with.contains(actor))
         {
             return false;
         }


rust/src/credential/activity_log.rs
```diff
@@ -103,14 +101,6 @@ impl ActivityLogFilterOptions {
             return false;
```
jessmelanson: This is unrelated to the max items changes, but on the verifier name filtering that I should've thought of before, sorry. Can the filtering check for containing the actor substring instead of equality?

spruceduncan (approved):

PR #258, opened 2025-12-19
[Flutter] DC API

Juliano1612:
## Description

This adds an example on how to add DC API support to a Flutter app using `sprucekit-mobile`.

### Other changes

Also adds an option to configure the app display name during the DC API presentation.

### Optional section

N/A

## Tested

- Tested the presentation with one internal Use Case

### DIFF
--- a/flutter/ios/Classes/CredentialPackAdapter.swift
+++ b/flutter/ios/Classes/CredentialPackAdapter.swift
@@ -9,26 +9,7 @@ class CredentialPackAdapter: CredentialPack {
 
     // In-memory store of credential packs
     private var packs: [String: SpruceIDMobileSdk.CredentialPack] = [:]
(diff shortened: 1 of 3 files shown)


flutter/ios/Classes/CredentialPackAdapter.swift
```diff
@@ -9,8 +9,27 @@ class CredentialPackAdapter: CredentialPack {
 
     // In-memory store of credential packs
     private var packs: [String: SpruceIDMobileSdk.CredentialPack] = [:]
+    // Store raw credentials for DC API sync (keyed by credentialId)
```
SeaSkyThe: Not sure if I got the reason behind creating a raw credential storage. Can't we get it from the credential pack?

PR #303, opened 2026-04-23
Load VDCs by ID and scope parameter for CredentialPack persistance

Juliano1612:
## Description

This adds scope isolation to `CredentialPack` persistence. The `scope` parameter lives in the native SDK (iOS + Android) and the Flutter adapters just pass `userHash` through. When `scope` is `null`, behavior is unchanged and existing storage keys stay intact.

Breaking changes
- Flutter plugin: Dart API signatures gained `userHash`. External consumers must pass `null` for legacy behavior.
- Native SDKs: source-compatible (default `nil/null`). Kotlin consumers need recompile due to JVM signature changes.

### Other changes

N/A

### Optional section

This PR exceeds 500 lines because it has generated code (>600 lines)

## Tested

Run the native app and test issuance and presentation


SeaSkyThe (approved):
In my understanding this doesn't harm apps that use the native versions of sprucekit, so I'm fine with it! Looks good.

Out of curiosity, lets see what Github Copilot says about the PR

PR #306, opened 2026-04-28
Add 18013-7 backwards compatible facade

spruceduncan:
This PR adds a backwards compatible facade for the 18013-7 OID4VP flow similar to the OID4VP and OID4VCI facades. It leaves the current v1 functionality in place, and just adds on additional functionalities, restoring the previous draft 18 flows.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -40,6 +40,7 @@ hpke = { version = "0.13.0", default-features = false, features = [
 ] }
 json-syntax = "0.12.5"
 itertools = "0.13"
+josekit = { git = "https://github.com/cobward/josekit-rs", rev = "635c8a7" }
 log = { version = "0.4", features = ["std", "serde"] }
 miniz_oxide = "0.7.2"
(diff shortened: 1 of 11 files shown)


rust/Cargo.toml
```diff
@@ -40,6 +40,7 @@ hpke = { version = "0.13.0", default-features = false, features = [
 ] }
 json-syntax = "0.12.5"
 itertools = "0.13"
+josekit = { git = "https://github.com/cobward/josekit-rs", rev = "635c8a7" }
```
w4ll3, line 43: why are we using this fork?

spruceduncan: @w4ll3 this is just the package we were using previously, this is the reason that was previously documented for using:

Patch of josekit to support RustCrypto for JWE generation in the 18013-7 Annex B OID4VP profile. Default josekit uses openssl which cannot be easily used in a mobile library.

Ryanmtate (approved):

PR #307, opened 2026-04-28
Flutter: OID4VP fixes

SeaSkyThe:
## Description

1. Credentials that answers different DCQL subqueries were being indexed with the same index:

    We were  identifying credentials by an int position. The sub-list APIs (`getCredentialRequirements`, `getCredentialsGroupedByQuery`) emitted positions local to each DCQL subquery, which then collided with the flat-list positions used by subsequent calls. Example below:
    
    Failure case: a DCQL credential_set with options=[[A], [B]] and a wallet holding one credential matching A and another matching B. Both land at position 0 of their respective option sub-list, so Dart sees two distinct credentials both labeled index=0. `getRequestedFields(0)` or `submitResponse([0])` then resolves to whichever credential the adapter cached first — typically the wrong one.
    
    BREAKING: `PresentableCredentialData` drops `index` and adds `credentialQueryId`; getRequestedFields and submitResponse take PresentableCredentialKey.

### Other changes

_Describe any minor or "drive-by" changes here._

### Optional section
- [X] Open a related PR in `safe-sdk`: https://github.com/spruceid/safe-sdk/pull/19

## Tested

- OID4VP in test app keeps working


flutter/android/src/main/kotlin/com/spruceid/sprucekit_mobile/Oid4vpAdapter.kt
```diff
@@ -266,13 +282,27 @@ internal class Oid4vpAdapter(
...
+            val creds = req.credentials.map { cred ->
+                val credId = cred.asParsedCredential().id()
+                val qid = req.credentialQueryIds.firstOrNull { qid ->
+                    keyMap.containsKey(
+                        PresentableCredentialKey(
+                            credentialId = credId,
+                            credentialQueryId = qid
+                        )
+                    )
+                } ?: req.credentialQueryIds.firstOrNull() ?: ""
```
Juliano1612: The fallback `?: req.credentialQueryIds.firstOrNull() ?: ""` silently constructs an unresolvable key if the Rust invariant ("at least one matching qid per credential") is ever violated. I'd rather replace the fallback with `error("...")`.

flutter/ios/Classes/Oid4vpAdapter.swift
```diff
@@ -274,14 +287,26 @@ class Oid4vpAdapter: Oid4vp {
...
+            // for which `(credentialId, qid)` exists in `keyMap`. Rust's
+            // invariant guarantees at least one such qid exists per cred.
+            let creds = req.credentials.map { cred -> PresentableCredentialData in
+                let credId = cred.asParsedCredential().id()
+                let qid = req.credentialQueryIds.first { qid in
+                    keyMap[PresentableCredentialKey(
+                        credentialId: credId,
+                        credentialQueryId: qid
+                    )] != nil
+                } ?? req.credentialQueryIds.first ?? ""
```
Juliano1612: Same as the Kotlin mirror

flutter/example/lib/demos/oid4vp_demo.dart
```diff
@@ -573,7 +578,7 @@ class _Oid4vpDemoState extends State<Oid4vpDemo> {
                 'Selective Disclosure: ${cred.selectiveDisclosable ? 'Yes' : 'No'}',
               ),
               trailing: const Icon(Icons.chevron_right),
-              onTap: () => _selectCredential(index),
+              onTap: () => _selectCredential(cred),
             ),
           );
         }),
```
Juliano1612, line 585: Now that the same credential can appear multiple times in the list (once per matching DCQL query), the user sees identical-looking "Credential 1" / "Credential 2" entries. Not a blocker, but consider adding `credentialQueryId` to the subtitle so it's obvious why the same `credentialId` shows up twice

flutter/android/src/main/kotlin/com/spruceid/sprucekit_mobile/Oid4vpAdapter.kt
Juliano1612: The fallback `?: req.credentialQueryIds.firstOrNull() ?: ""` silently constructs an unresolvable key if the Rust invariant ("at least one matching qid per credential") is ever violated. I'd rather replace the fallback with an error.

Juliano1612 (approved):
LGTM! just two optional nits

PR #311, opened 2026-05-07
Add NFC reader support in Android

sbihel:
Tested with:
- Samsung Wallet
- Google Wallet
- Showcase

Remaining issues:
- [X] the reader seems to keep sending the initial APDU after the engagement is complete, so e.g. on a Samsung phone the wallet picker will pop up again when reviewing the presentation reques
- [X] the NFC engagement is very finicky, but I don't think it's about the engagement:
    * between the google pixel and the samsung phone, I believe there are conflicts with the payment protocols and the mDL presentation doesn't seem to suppress them. If you discard warnings about tag types, you will find that the presentation is still ongoing underneath
- [x] Samsung reader: the first tap with Google Pixel still has "unknown tag" screen but it doesn't happen on the second tap anymore
- [x] Google Pixel reader: Samsung still complains about payment NFC conflicts on the first tap, but then it's all good
    * Turns out it isn't about payment apps, it's just the OS saying the wallet picker can be avoided if the user disables wallets they're not using. And anyway it's for the wallet side, it's not caused by the reader.
Screenshot_20260512_141938_One UI Home

- [x] The code (and UI) was fully AI generated, and hasn't yet been reviewed.

### DIFF
--- a/android/MobileSdk/src/main/AndroidManifest.xml
+++ b/android/MobileSdk/src/main/AndroidManifest.xml
@@ -17,4 +17,9 @@
     <uses-permission android:name="android.permission.CAMERA" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.INTERNET"/>
+    <uses-permission android:name="android.permission.NFC" />
+
+    <uses-feature
+        android:name="android.hardware.nfc"
+        android:required="false" />
 </manifest>
--- a/android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/IsoMdlReader.kt
+++ b/android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/IsoMdlReader.kt
@@ -15,7 +15,7 @@ import java.util.UUID
 
 class IsoMdlReader(
     val callback: BLESessionStateDelegate,
-    uri: String,
+    handover: ReaderHandover,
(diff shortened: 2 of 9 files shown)


tjzhang-spruce:
Didn't observe outstanding issue from my side but AI bot had some callouts I think could be useful:

```
1. High Priority: Thread Safety in readerCallback
The onTagDiscovered callback (via readerCallback) runs on the NFC controller's background thread.

Issue: The active flag (even with @Volatile) is checked and set in a non-atomic way. If a user "double-taps" or the tag is detected multiple times in rapid succession, two concurrent APDU exchange flows could trigger.

Risk: This can lead to protocol desynchronization or "Transceive failed" errors that are hard to debug.

Recommendation: Use an AtomicBoolean with compareAndSet to ensure only one session is active at a time.

2. Medium Priority: Potential Memory Leak in CameraAvailabilityCallback
Issue: The cameraAvailabilityCallback is registered as an anonymous inner class to the system-level CameraManager.

Risk: If the NfcReaderEngagement is not properly released (e.g., during an Activity configuration change or an unexpected lifecycle termination), the CameraManager will keep a strong reference to the callback, which in turn holds the Activity context, causing a memory leak.

Recommendation: Ensure unregisterAvailabilityCallback is called in all teardown paths, or use a WeakReference for context-dependent logic.
```

tjzhang-spruce (approved):

PR #322, opened 2026-06-12
Bump isomdl for NFC negotiated-handover fixes and honor TNEP pacing

tjzhang-spruce:
## Summary

Picks up [isomdl#130](https://github.com/spruceid/isomdl/pull/130) (rev `a4f92cc` → `fb19e7b`), which fixes reading Apple Wallet over negotiated NFC handover, and wires its new pacing hint into both NFC reader engagement loops.

## What the isomdl bump brings

- **Apple Wallet infinite-loop fix**: an empty NDEF message (NLEN=0) during negotiated handover is TNEP's "response not ready" signal; the driver now polls with a bounded budget instead of looping until the NFC link dies.
- **Chunk size negotiated from the holder's CC MLe** (and a fixed Le encoding that previously truncated 264 → 8).
- **DeviceEngagement version 1.1 accepted** (ISO 18013-5 Second Edition — what Apple emits), with the spec's version/key-5-6 consistency rule enforced.
- **`recommendedDelayMs()`** — additive driver method exposing TNEP's minimum waiting time (parsed from the holder's Tp record; Apple advertises 16 ms).

## Changes here

- `rust/Cargo.toml` + `Cargo.lock`: isomdl rev bump.
- `rust/src/mdl/reader.rs`: expose `recommended_delay_ms()` through the UniFFI wrapper (additive; `ReaderApduProgress` unchanged, so no caller breaks).
- `android/.../NfcReaderEngagement.kt` + `ios/.../NfcReaderEngagement.swift`: sleep the recommended delay before the next transceive, so polling doesn't compete with the holder preparing its Handover Select.
- `rust/MobileSdkRs/.../mobile_sdk_rs.swift`: regenerated bindings.

## Verification

- `cargo build` with `-Dwarnings` — clean.
- Android `:mobilesdk:compileReleaseKotlin` — BUILD SUCCESSFUL.
- iOS `xcodebuild -scheme MobileSdk` (simulator) — BUILD SUCCEEDED.
- Real-device regression pending: Pixel reader × {Apple, Google, Samsung} Wallet via the native Showcase verifier; will report results on this PR.

### DIFF
--- a/ios/MobileSdk/Sources/MobileSdk/nfc/NfcReaderEngagement.swift
+++ b/ios/MobileSdk/Sources/MobileSdk/nfc/NfcReaderEngagement.swift
@@ -182,6 +182,7 @@ extension NfcReaderEngagement: NFCTagReaderSessionDelegate {
                     // preparing its Handover Select).
                     let delayMs = driverInit.driver.recommendedDelayMs()
                     if delayMs > 0 {
+                        // Task.sleep(for:) needs iOS 16; the SDK targets 14.0.
                         try await Task.sleep(nanoseconds: UInt64(delayMs) * 1_000_000)
                     }
                     rapdu = try await sendCommand(nextApdu, on: tag)
--- a/ios/MobileSdk/Sources/MobileSdk/nfc/NfcReaderEngagement.swift
+++ b/ios/MobileSdk/Sources/MobileSdk/nfc/NfcReaderEngagement.swift
@@ -177,6 +177,13 @@ extension NfcReaderEngagement: NFCTagReaderSessionDelegate {
                 let progress = try driverInit.driver.processRapdu(command: rapdu)
                 switch progress {
                 case .inProgress(let nextApdu):
+                    // TNEP minimum waiting time; 0 unless the holder asked us
+                    // to pause before the next read (e.g. Apple Wallet
+                    // preparing its Handover Select).
+                    let delayMs = driverInit.driver.recommendedDelayMs()
+                    if delayMs > 0 {
+                        try await Task.sleep(nanoseconds: UInt64(delayMs) * 1_000_000)
+                    }
                     rapdu = try await sendCommand(nextApdu, on: tag)
                 case .done(let handover):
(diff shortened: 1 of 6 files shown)


ios/MobileSdk/Sources/MobileSdk/nfc/NfcReaderEngagement.swift
```diff
@@ -177,6 +177,13 @@ extension NfcReaderEngagement: NFCTagReaderSessionDelegate {
                 let progress = try driverInit.driver.processRapdu(command: rapdu)
                 switch progress {
                 case .inProgress(let nextApdu):
+                    // TNEP minimum waiting time; 0 unless the holder asked us
+                    // to pause before the next read (e.g. Apple Wallet
+                    // preparing its Handover Select).
+                    let delayMs = driverInit.driver.recommendedDelayMs()
+                    if delayMs > 0 {
+                        try await Task.sleep(nanoseconds: UInt64(delayMs) * 1_000_000)
```
w4ll3, line 186: can't you just `Task.sleep(for: .milliseconds(delayMs))`?

tjzhang-spruce: Tried it — doesn't compile: Duration/Task.sleep(for:) is iOS 16+, and the SDK's deployment target is 14.0 (Package.swift pins .iOS(.v14)).

```
error: 'sleep(for:tolerance:clock:)' is only available in iOS 16.0 or newer
error: 'milliseconds' is only available in iOS 16.0 or newer
```

Ryanmtate (approved):

ios/MobileSdk/Sources/MobileSdk/nfc/NfcReaderEngagement.swift
w4ll3: Mmmm, I thought we were 16+ already, I'll bring up this dicussion with a broader group but I do think we should upgrade.

w4ll3 (approved):

PR #131, opened 2025-03-20
Wallet Service Client

Juliano1612:
## Description

This implements a WalletServiceClient to authenticate the showcase app. 

### Optional section

This PR exceeds 500 lines because it contains generated swift code.

## Tested

Unit tests were implemented using a [Wiremock](https://docs.rs/wiremock/0.6.3/wiremock/) server

PR #323, opened 2026-06-12
Fix GattServer rejecting the final BLE data chunk; add reader diagnostics

tjzhang-spruce:
## Summary

Found while testing the NFC mDL reader against an Apple Wallet holder (Google/Samsung work). Two commits:

1. **Fix** a logic bug in `GattServer` that flagged the final Client2Server data chunk as invalid.
2. **Diagnostics** to pin down a still-open Apple-specific empty-result issue (see "What this does NOT fix" below).

## Commit 1 — final-chunk handling

Per ISO 18013-5 the data characteristics chunk messages with a leading byte: `0x00` = final chunk, `0x01` = intermediate. The handler assembled and delivered the message correctly on `0x00`, but the validation right after was shaped as:

```kotlin
if (value[0] == 0x01) { /* size check */ } else { onError("Invalid first byte … expected 0 or 1") }
```

so a `0x00` byte fell into the `else` and raised **"Invalid first byte 0 … expected 0 or 1"** — self-contradictory, since `0` is one of the expected values. That spurious `onError` drove the session to `ERROR` and tripped a teardown-race `NullPointerException` in `onNotificationSent` (the GATT server was already nulled). All of this fired *after* the message had been delivered, so it was latent for holders whose response was captured first, but it's a real bug.

Replaced with a three-way `when`: `0x00` is a no-op (message already delivered), `0x01` keeps its MTU size check, only other bytes error.

Observed in a real Pixel-reader → Apple-Wallet log:
```
TransportBlePeripheralServerReader: Invalid first byte 0 in Client2Server data chunk, expected 0 or 1
BluetoothGattServer: Unhandled exception: java.lang.NullPointerException: … onNotificationSent … on a null object reference
```

## Commit 2 — diagnostics

The same log showed the NFC handover + isomdl fixes working end to end (polling, TNEP pacing, MLe negotiation, DeviceEngagement 1.1), the BLE session connecting, and the reader's request going out — but the Apple result screen showed no fields with both issuer and device auth `Unchecked`. The response *was* received (`onMessageReceived` fired), so the empty result is **not** explained by commit 1.

To localize that, this commit logs:
- the received BLE message size + hex prefix (the previous `data.toString()` only printed an object hash), and
- `handleResponse`'s outcome — doc types, issuer/device auth status, namespace keys, and the isomdl error map.

That will distinguish "short/empty response over BLE" from "full response that failed to decrypt or parse" (leading hypothesis: NFC-handover SessionTranscript mismatch between reader and Apple).

## What this does NOT fix

The Apple Wallet empty-result issue itself is **not** resolved here — commit 1 removes a spurious error + NPE, commit 2 adds the logging needed to root-cause it. Next step is a fresh capture from the Apple holder with these diagnostics.

## Verification

- `:mobilesdk:compileReleaseKotlin` — BUILD SUCCESSFUL.
- Real-device regression (Google/Samsung still read cleanly; Apple capture with new diagnostics) — to follow on this PR.

### DIFF
--- a/android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/IsoMdlReader.kt
+++ b/android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/IsoMdlReader.kt
@@ -111,8 +111,21 @@ class IsoMdlReader(
 
     fun handleMdlReaderResponseData(response: ByteArray): MdlReaderResponseData {
         try {
-            return com.spruceid.mobile.sdk.rs.handleResponse(session, response)
+            val data = com.spruceid.mobile.sdk.rs.handleResponse(session, response)
+            // Diagnostic: surface what handleResponse produced so a capture can
+            // tell "empty/failed parse" from "parsed but unverified". `errors`
+            // is the JSON-encoded per-category error map from isomdl.
+            Log.d(
+                "IsoMdlReader",
+                "handleResponse: docTypes=${data.docTypes}, " +
+                    "issuerAuth=${data.issuerAuthentication}, " +
+                    "deviceAuth=${data.deviceAuthentication}, " +
+                    "namespaces=${data.verifiedResponse.keys}, " +
+                    "errors=${data.errors}"
+            )
+            return data
         } catch (e: MdlReaderResponseException) {
+            Log.e("IsoMdlReader", "handleResponse failed", e)
             throw e
         }
     }
--- a/android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/ble/GattServer.kt
+++ b/android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/ble/GattServer.kt
@@ -324,32 +324,39 @@ class GattServer(
                         }
                     } // End synchronized block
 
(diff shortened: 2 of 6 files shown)


tjzhang-spruce:
Was able to perform successful presentation for Apple Google Samsung wallet with Bryce today. 
[nfc-apple-test (2).log](https://github.com/user-attachments/files/28904904/nfc-apple-test.2.log)

sbihel (approved):
Would be nice to have some tests for this type of logic 🙈

tjzhang-spruce:
> Would be nice to have some tests for this type of logic 🙈

Re-pinned isomdl to the merged main SHA (3ad4471, [#131](https://github.com/spruceid/isomdl/issues/131)). And addressed the test ask: extracted the Client2Server chunk framing into a pure classifyClient2ServerChunk() and added GattServerTest — 5 cases including the 0x00 final-chunk regression (must be accepted, not flagged "expected 0 or 1"), plus over-MTU and bad-first-byte. :MobileSdk:testDebugUnitTest green. Behavior unchanged.

PR #326, opened 2026-06-15
Fixes CD tag ordering and publish iOS pods to CocoaPods

SeaSkyThe:
## Description
 The release is tagged inside the ios job, which is just where `git tag` runs. The Flutter version bump ran in a job that executed after that, so the bump was not in the tagged commit: checking out the release tag returned stale Flutter versions (`pubspec.yaml`, `build.gradle`). Move the flutter job before ios so the bump is part of the tagged commit.

### Other changes

Also publish the iOS SDK pods to the private [spruceid/cocoapods-specs](https://github.com/spruceid/cocoapods-specs) repo after tagging, and move the `flutter/ios podspec` bump into the flutter job so all Flutter version edits live in one place. 

To publish to cocoapods, I followed: https://guides.cocoapods.org/making/private-cocoapods.html

### Optional section

- [x] Need to setup COCOAPODS_SPECS_DEPLOY_KEY so this repo can publish to: [spruceid/cocoapods-specs](https://github.com/spruceid/cocoapods-specs)

## Tested

- [x] Need to publish a test tag and test locally

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -81,7 +81,7 @@ jobs:
 
   ios:
     runs-on: macos-latest
-    needs: validation
+    needs: flutter
     permissions:
       contents: write
     steps:
@@ -105,6 +105,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           token: ${{ steps.app-token.outputs.token }}
+          ref: main
       - uses: dtolnay/rust-toolchain@stable
         with:
           targets: "aarch64-apple-ios-sim,aarch64-apple-ios,x86_64-apple-ios"
@@ -141,15 +142,13 @@ jobs:
           sed -i '' "s/spec.version.*=.*/spec.version      = '${{ github.event.inputs.version }}'/" SpruceIDMobileSdkRs.podspec
           sed -i '' "s/spec.version.*=.*/spec.version          = '${{ github.event.inputs.version }}'/" SpruceIDMobileSdk.podspec
           sed -i '' "s/spec.dependency 'SpruceIDMobileSdkRs'.*/spec.dependency 'SpruceIDMobileSdkRs', '~> ${{ github.event.inputs.version }}'/" SpruceIDMobileSdk.podspec
-          sed -i '' "s/s.version.*/s.version          = '${{ github.event.inputs.version }}'/" flutter/ios/sprucekit_mobile.podspec
-          sed -i '' "s/s.dependency 'SpruceIDMobileSdk'.*/s.dependency 'SpruceIDMobileSdk', '~> ${{ github.event.inputs.version }}'/" flutter/ios/sprucekit_mobile.podspec
 
       # If this fails due to a new commit on the main branch, it's a feature not a bug!
       # This way, there will be no confusion around what was included in the release.
       - name: Push changes and tag
         id: commit
         run: |
-          git add Package.swift SpruceIDMobileSdk.podspec SpruceIDMobileSdkRs.podspec SpruceIDMobileSdkRsRustFramework.podspec flutter/ios/sprucekit_mobile.podspec
+          git add Package.swift SpruceIDMobileSdk.podspec SpruceIDMobileSdkRs.podspec SpruceIDMobileSdkRsRustFramework.podspec
           git commit -m "Release ${{ github.event.inputs.version }}"
           git push
           git tag ${{ github.event.inputs.version }} -m "${{ github.event.inputs.version }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/cd.yml
```diff
@@ -162,6 +161,19 @@ jobs:
...
+      - uses: webfactory/ssh-agent@v0.9.0
+        if: steps.commit.outcome == 'success'
+        with:
+          ssh-private-key: ${{ secrets.COCOAPODS_SPECS_DEPLOY_KEY }}
+
+      - name: Publish podspecs to private Specs repo
+        if: steps.commit.outcome == 'success'
+        run: |
+          pod repo add spruce-specs git@github.com:spruceid/cocoapods-specs.git
+          pod repo push spruce-specs SpruceIDMobileSdkRsRustFramework.podspec --allow-warnings --skip-import-validation
```
sbihel, line 202: JFYI maybe you don't need `--skip-import-validation` as there won't be a CDN involved, but when I tried to use it last time I worked with cocoapods it was working. There are other flags like `--skip-tests` or `--quick` that may help but i didn't have much luck

.github/workflows/cd.yml
```diff
@@ -81,7 +81,7 @@ jobs:
 
   ios:
     runs-on: macos-latest
-    needs: validation
+    needs: flutter
```
sbihel: If the `ios` and `android` jobs don't run in parallel it will effectively double the execution time of the cd action. Do you think there's a way of extracting the git commit/tag to its own job so we can keep them in parallel?

sbihel (commented):
This is a bit separate but it looks like the `Revert Package.swift for local development` step might need some fixing as well. Release 1.70.0 failed, and now it doesn't properly change the binary dependency back to a local path (see https://github.com/spruceid/sprucekit-mobile/commit/08a16292a17cf5dbab738a08f8e9025afc1c66ed). Maybe instead of a git revert it should simply do a `sed`

.github/workflows/cd.yml
SeaSkyThe: - `validation` runs first
- `ios` and `android` in parallel
    - `ios` builds and creates an artifact to be used later
- `release` is a new step, that bumps all the tags in iOS and Flutter files.

`release` step does, in order:
1. Check main didn't move since dispatch (else stop) -- keeps tag source == built binary.
2. Download the iOS zip created in the `ios` step
3. Bump version everywhere: Package.swift, 3 iOS podspecs, Flutter files (flutter is now inside the release process)
4. Commit Release 0.17.0 → push → tag → push tag.
5. Create GitHub Release + attach zip.
6. Publish pods to private Specs repo.
7. Revert Package.swift to local-dev

Also change the check to make the Package.swift revert run even if the GitHub release or pod push fails, so main is always restored to the local-dev binary path.  

Whats your suggestion to better test it? Merge and create a random tag? Like 1.17.0-test.0.0.1

sbihel: i think it looks good now, but it's always hard to review without a real run.

You can trigger the cd action and point at this branch, this should work fine. For the tag, using semver extension might not work (e.g. i don't think it's supported by Maven) so I would recommend using something like 0.0.20

Also make sure to fix the current https://github.com/spruceid/sprucekit-mobile/blob/1435071db3f99546d8098a88c31618c4bdd306cd/Package.swift#L24 to be fit for local development

SeaSkyThe:
It worked, here is the test CD: https://github.com/spruceid/sprucekit-mobile/actions/runs/27633477250.

You can see the 0.0.50 version published at: https://github.com/spruceid/cocoapods-specs

I already reverted the changes the CD makes in the branch files, but everything went ok

.github/workflows/cd.yml
```diff
@@ -105,51 +140,44 @@ jobs:
...
-
-      - name: Generate Swift package
-        run: cargo swift package -p ios -n MobileSdkRs --release
-        working-directory: rust
+          ref: main
 
-      - name: Compress XCFramework
+      # If this fails due to a new commit on the main branch, it's a feature not a bug!
+      # This way, there will be no confusion around what was included in the release.
+      - name: Ensure main has not moved since dispatch
```
sbihel, line 147: Is this better than staying on the commit used for previous steps?

sbihel (approved):
I'm assuming you've already deleted the release/tag?

And did you double check that things work well with a Flutter app?

.github/workflows/cd.yml
SeaSkyThe: > I'm assuming you've already deleted the release/tag?

Yes

> And did you double check that things work well with a Flutter app?

I was able to resolve/download the pods from the private repo, so I'm assuming it will work just fine.

> Is this better than staying on the commit used for previous steps?

I honestly just tried maintain the same behaviour we had before, but with an explicit check instead of a implicit one. Because even if we try to stay in the commit used for previous steps, and someone pushes to main, our push (related to tag update and seds) will fail anyway, unless we force push, which I don't think is safe

PR #330, opened 2026-06-23
WIP: support retaining expired credentials for DC-API

alichay:
Allow control of iOS's `invalidationDate` for DC-API, so the matcher doesn't filter expired credentials

All behavior changes should be opt-in

### DIFF
--- a/ios/MobileSdk/Sources/MobileSdk/CredentialPack.swift
+++ b/ios/MobileSdk/Sources/MobileSdk/CredentialPack.swift
@@ -216,14 +216,6 @@ public class CredentialPack {
                         documentIdentifier: mdoc.id(),
                         invalidationDate: invalidationDate
                     )
-                    print("""
-                    🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨
-                    🚨 GOT HERE: addMDocToIDProvider → addRegistration   🚨
-                    🚨 mdoc.id() = \(mdoc.id())
-                    🚨 policy = \(invalidationPolicy)
-                    🚨 invalidationDate FORCED TO nil (TEMPORARY TEST HACK) 🚨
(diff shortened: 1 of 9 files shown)


ios/MobileSdk/Sources/MobileSdk/CredentialPack.swift
```diff
@@ -148,45 +166,81 @@ public class CredentialPack {
...
+                        for: mdoc,
+                        policy: invalidationPolicy
+                    )
+                    let registration = MobileDocumentRegistration(
+                        mobileDocumentType: "org.iso.18013.5.1.mDL",
+                        supportedAuthorityKeyIdentifiers: [],  // TODO
+                        documentIdentifier: mdoc.id(),
+                        invalidationDate: invalidationDate
+                    )
+                    print("""
```
tjzhang-spruce: Seems to be for testing purpose, we can remove it before merging

alichay: I manually staged changes before pushing, I have no idea how this ended up pushed 😭 

Fixed!

ios/Showcase/Targets/AppUIKit/Sources/wallet/HandleMdocOID4VPView.swift
```diff
@@ -179,6 +179,9 @@ struct MdocFieldSelector: View {
     @State private var selectedFields: [String]
     let requiredFields: [String]
 
+    /// Set by the user to share a credential that is expired / not yet valid.
+    @State private var acknowledgedInvalid = false
```
tjzhang-spruce, line 184: Since this is a local state, if a user ever has multiple invalid credentials and switches between them in the UI, the "Share anyway" checkbox might remain auto-checked from the previous credential (bypassing the safety warning).
Given that most users likely only have one credential right now, this is definitely an edge case and not a blocker. However, for future-proofing, it might be worth simply binding the view to .id(match.id) or resetting the state when the credential changes. Feel free to treat this as a non-blocking nitpick!

tjzhang-spruce (approved):

Juliano1612 (approved):
LGTM!

PR #332, opened 2026-06-24
Continuously observes network reachability and keeps `hasConnection` up to date.

patmalt:
## Description

This PR replaces the blocking, on-demand internet connectivity check on the iOS Showcase with a continuous network reachability observer.

Previously, checkInternetConnection() (in Utils.swift) spun up a new `NWPathMonitor` on every call and used a `DispatchSemaphore` to block the calling thread until the first path update arrived, then synchronously returned the result. This had to be invoked manually at every point where connectivity mattered (onAppear, pull-to-refresh, and after generating a mock mDL), and the blocking semaphore wait risked stalling whatever thread it was called from.

The updated behavior moves connectivity tracking into `StatusListObservable`. A single long-lived `NWPathMonitor` runs on a dedicated background queue and updates the published `hasConnection` property whenever the network path changes, hopping to the main actor for the assignment so SwiftUI observers update safely. The monitor is started in init and cancelled in deinit. As a result, `hasConnection` now stays continuously up to date on its own, and all the manual checkInternetConnection() call sites in WalletHomeView are removed.

This removes the semaphore-based blocking pattern entirely and ensures connectivity state reflects real-time changes rather than a one-shot snapshot taken at specific UI events.

### Other changes

- Made hasConnection private(set) since it is now owned and updated exclusively by StatusListObservable rather than mutated externally from the views.
- Deleted the now-unused checkInternetConnection() helper and dropped the import Network from Utils.swift, adding it to StatusListObservable.swift where it is now needed.

## Tested

1. Launched the app and confirmed connectivity-dependent UI reflects the current network state on the wallet home screen.
2. Toggled the device/simulator network on and off (e.g. via Airplane Mode) and verified hasConnection updates automatically without needing to pull-to-refresh or re-enter the screen.
3. Exercised pull-to-refresh, mock mDL generation, and initial screen load to confirm behavior is unchanged now that the explicit connectivity checks have been removed.

### DIFF
--- a/ios/Showcase/Targets/AppUIKit/Sources/Utils.swift
+++ b/ios/Showcase/Targets/AppUIKit/Sources/Utils.swift
@@ -3,7 +3,6 @@ import SpruceIDMobileSdk
 import SpruceIDMobileSdkRs
 import Foundation
 import SwiftUI
-import Network
 
 // modifier
 struct HideViewModifier: ViewModifier {
@@ -88,25 +87,6 @@ func generateQRCode(from data: Data) -> UIImage {
         }
     }
     return UIImage(systemName: "xmark.circle") ?? UIImage()
-}
-
-func checkInternetConnection() -> Bool {
-    let monitor = NWPathMonitor()
-    let queue = DispatchQueue.global(qos: .background)
-    var isConnected = false
-
-    let semaphore = DispatchSemaphore(value: 0)
-
-    monitor.pathUpdateHandler = { path in
-        isConnected = (path.status == .satisfied)
-        semaphore.signal()
-        monitor.cancel()
-    }
-
-    monitor.start(queue: queue)
-    semaphore.wait()
-
-    return isConnected
 }
 
 func generateTxtFile(content: String, filename: String) -> URL? {
--- a/ios/Showcase/Targets/AppUIKit/Sources/dataStores/StatusListObservable.swift
+++ b/ios/Showcase/Targets/AppUIKit/Sources/dataStores/StatusListObservable.swift
@@ -1,9 +1,14 @@
 import Foundation
+import Network
 import SpruceIDMobileSdk
 
 class StatusListObservable: ObservableObject {
     @Published var statusLists: [String: CredentialStatusList]
-    @Published var hasConnection: Bool
+    @Published private(set) var hasConnection: Bool
+
+    private let connectionMonitor = NWPathMonitor()
+    private let connectionMonitorQueue = DispatchQueue(
+        label: "com.spruceid.showcase.StatusListObservable.connectionMonitor")
 
     init(
         statusLists: [String: CredentialStatusList] = [:],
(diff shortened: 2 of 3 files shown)


SeaSkyThe (approved):
Looks good!

Just to make sure I understood correctly: we're only changing how we check for the connection, so this shouldn't impact any features that depend on it. The main difference is that the connection check is now done asynchronously, thats correct?

patmalt:
> Looks good!
> 
> Just to make sure I understood correctly: we're only changing how we check for the connection, so this shouldn't impact any features that depend on it. The main difference is that the connection check is now done asynchronously, thats correct?

Previously, `StatusListObservable.hasConnection` was set from the result of calling `checkInternetConnection()`.  `checkInternetConnection` used `DispatchSemaphore` and `semaphore.wait()`, even on the main thread, causing the main thread to become blocked momentarily. Additionally, the QoS of `.background` for the notification queue of `NWPathMonitor` caused runtime warnings indicating the possibility of priority inversion. Priority inversion can cause issues where the caller of `checkInternetConnection` is of a higher QoS than `.background`.

Now, `NWPathMonitor` is a part of `StatusListObservable`, the monitoring is created on initialization and continues for the lifetime of `StatusListObservable`. `StatusListObservable` is an environment object of `ContentView`, the body of the app. Therefore, the lifetime of our observation of networking will continue for the entire lifetime of the app. Many benefits here, including reducing the allocation of transient `NWPathMonitor` instances and quicker look up time when the network connection has not changed.

One down side is the list is optimistic, having an initial value for `hasConnection` is `true`. Since we know `NWPathMonitor` is relatively quick, quick enough to currently block the main thread, the optimism should not be harmful. @SeaSkyThe

SeaSkyThe:
> > Looks good!
> > Just to make sure I understood correctly: we're only changing how we check for the connection, so this shouldn't impact any features that depend on it. The main difference is that the connection check is now done asynchronously, thats correct?
> 
> Previously, `StatusListObservable.hasConnection` was set from the result of calling `checkInternetConnection()`. `checkInternetConnection` used `DispatchSemaphore` and `semaphore.wait()`, even on the main thread, causing the main thread to become blocked momentarily. Additionally, the QoS of `.background` for the notification queue of `NWPathMonitor` caused runtime warnings indicating the possibility of priority inversion. Priority inversion can cause issues where the caller of `checkInternetConnection` is of a higher QoS than `.background`.
> 
> Now, `NWPathMonitor` is a part of `StatusListObservable`, the monitoring is created on initialization and continues for the lifetime of `StatusListObservable`. `StatusListObservable` is an environment object of `ContentView`, the body of the app. Therefore, the lifetime of our observation of networking will continue for the entire lifetime of the app. Many benefits here, including reducing the allocation of transient `NWPathMonitor` instances and quicker look up time when the network connection has not changed.
> 
> One down side is the list is optimistic, having an initial value for `hasConnection` is `true`. Since we know `NWPathMonitor` is relatively quick, quick enough to currently block the main thread, the optimism should not be harmful. @SeaSkyThe

Lovely! Thanks for the explanation. Feel free to merge it whenever!

PR #335, opened 2026-06-30
iOS holder: NFC static-handover presentation (CAMV-975)

tjzhang-spruce:
## Summary
Mirror the reader's `fromHandover:` entry on the holder side so the iOS SDK can present an mDL from an **NFC static-handover** engagement, not just a QR one.

`MdocProximityPresentationManager` gains `DeviceEngagement.NFC(NegotiatedCarrierInfo)`: it starts a BLE **central-client-mode** session on the UUID fixed by the handover and reuses the existing BLE transport + session code (the same path as QR). The Rust `ApduHandoverDriver` / `DeviceEngagementData.NFC` primitives already existed and are unchanged — this is the missing Swift wiring on top.

Linear: CAMV-975 (Apple NFC & SE Platform).

## ⚠️ Breaking change — release as 0.18.x
Adds two **public enum cases**: `DeviceEngagement.NFC(...)` and `MdocProximityPresentationManager.State.connectingViaNfc`. Downstream code that does an **exhaustive `switch` without a `default`** over either enum (e.g. a consumer's `connectionState(changedTo:)` handler, or the wallet) must add a branch to compile. No init/signature changes — call sites that don't switch over these enums are unaffected.

## Changes
- **PresentationManager**: `.NFC(carrier)` engagement + `State.connectingViaNfc`; `reset()`/`setupSession()` route NFC to central-only on the carrier UUID, skipping QR generation. Logs a warning if an `.NFC` engagement is passed no central transmission mode.
- **CentralManager**: mdoc `init` that scans a pre-negotiated service UUID (instead of a freshly generated one).
- **MdlPresentationAdapter** (Flutter plugin): handle the new state — unreachable via the plugin's pigeon today, keeps the switch exhaustive.
- **Showcase**: new **NFC tab** (matching Android's Share-NFC tab — `Wallet` icon, same slot) hosting a single-device **wiring self-test**: an in-memory APDU loopback (`ApduHandoverDriver` ↔ `ReaderApduHandoverDriver`) brings up both BLE roles to verify the NFC→BLE wiring **without NFC hardware or the SE entitlement**.

## Testing
- iOS SDK + Showcase build green; verified on a physical device (iPhone, iOS 26.5.1): the NFC tab self-test reaches **"Wiring verified"** (holder `connectingViaNfc` + reader `connecting`).
- A single phone can't BLE-connect to itself, so the self-test verifies wiring only. The full BLE session + response is the **same code already exercised by the QR flow**. A true over-the-air tap is gated on the Apple SE entitlement + the Java Card applet (out of scope here).
- No Rust changes; generated bindings unchanged (CI `git diff --exit-code` passes).

### DIFF
--- a/ios/MobileSdk/Sources/MobileSdk/mdoc/proximity/PresentationManager.swift
+++ b/ios/MobileSdk/Sources/MobileSdk/mdoc/proximity/PresentationManager.swift
@@ -61,11 +61,16 @@ public class MdocProximityPresentationManager {
         case let .NFC(carrier):
             // NFC static handover is BLE central-client-mode only, on the UUID fixed by the handover.
             var l2capUsage: L2CAPUsage = .disableL2CAP
+            var hasCentralMode = false
             for mode in transmissionModes {
                 if case let .bleMdocCentralMode(usage) = mode {
                     l2capUsage = usage
+                    hasCentralMode = true
                     break
                 }
+            }
+            if !hasCentralMode {
+                print("NFC engagement forces BLE mdoc central client mode; ignoring the provided transmission modes")
             }
             central = CentralManager(mdoc: handle, serviceUuid: carrier.getUuid(), l2capUsage)
         }
--- a/ios/MobileSdk/Sources/MobileSdk/mdoc/proximity/PresentationManager.swift
+++ b/ios/MobileSdk/Sources/MobileSdk/mdoc/proximity/PresentationManager.swift
@@ -38,23 +38,36 @@ public class MdocProximityPresentationManager {
 
         var central: CentralManager?
         var peripheral: PeripheralManager?
-        let handle = DelegateWrapper(delegate: delegate, mdoc: mdoc)
+        let handle = DelegateWrapper(delegate: delegate, mdoc: mdoc, engagement: engagement)
 
-        for mode in transmissionModes {
-            switch mode {
-            case let .bleMdocCentralMode(l2capUsage):
-                if central != nil {
-                    print("multiple .bleMdocCentralMode instances provided, ignoring all but the first")
-                    break
+        switch engagement {
+        case .QRCode:
+            for mode in transmissionModes {
+                switch mode {
+                case let .bleMdocCentralMode(l2capUsage):
+                    if central != nil {
+                        print("multiple .bleMdocCentralMode instances provided, ignoring all but the first")
+                        break
+                    }
+                    central = CentralManager(mdoc: handle, l2capUsage)
+                case let .bleMdocPeripheralMode(l2capUsage):
+                    if peripheral != nil {
+                        print("multiple .bleMdocPeripheralMode instances provided, ignoring all but the first")
+                        break
+                    }
+                    peripheral = PeripheralManager(mdoc: handle, l2capUsage)
                 }
-                central = CentralManager(mdoc: handle, l2capUsage)
-            case let .bleMdocPeripheralMode(l2capUsage):
-                if peripheral != nil {
-                    print("multiple .bleMdocPeripheralMode instances provided, ignoring all but the first")
-                    break
-                }
-                peripheral = PeripheralManager(mdoc: handle, l2capUsage)
             }
+        case let .NFC(carrier):
+            // NFC static handover is BLE central-client-mode only, on the UUID fixed by the handover.
+            var l2capUsage: L2CAPUsage = .disableL2CAP
+            for mode in transmissionModes {
+                if case let .bleMdocCentralMode(usage) = mode {
+                    l2capUsage = usage
+                    break
+                }
+            }
+            central = CentralManager(mdoc: handle, serviceUuid: carrier.getUuid(), l2capUsage)
         }
 
         handle.set(bleCentral: central)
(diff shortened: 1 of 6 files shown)


ios/MobileSdk/Sources/MobileSdk/mdoc/proximity/PresentationManager.swift
```diff
@@ -367,6 +398,8 @@ public class MdocProximityPresentationManager {
         case action(required: RequiredAction)
         /// Attempting to establish a connection with the reader.
         case connecting(qrPayload: Data)
+        /// Awaiting the reader; engagement was delivered over NFC, so there's no QR payload.
+        case connectingViaNfc
```
Juliano1612, line 407: Just a heads up to release this as 0.18.x, as this is a breaking change

ios/MobileSdk/Sources/MobileSdk/mdoc/proximity/PresentationManager.swift
```diff
@@ -38,23 +38,36 @@ public class MdocProximityPresentationManager {
...
+        case let .NFC(carrier):
+            // NFC static handover is BLE central-client-mode only, on the UUID fixed by the handover.
+            var l2capUsage: L2CAPUsage = .disableL2CAP
+            for mode in transmissionModes {
+                if case let .bleMdocCentralMode(usage) = mode {
+                    l2capUsage = usage
                     break
                 }
-                peripheral = PeripheralManager(mdoc: handle, l2capUsage)
             }
```
Juliano1612, line 71: In `.NFC`, you always create a `CentralManager` and only scan `transmissionModes` for an L2CAP setting. If a caller passes only `.bleMdocPeripheralMode`, that's silently ignored and central is created anyway with `disableL2CAP`. So this mirrors the QR branch's warnings but here there's none. Can you print a warning when no central mode is present (or document that NFC forces central-client mode)?

ios/Showcase/Targets/AppUIKit/Sources/credentials/ShareMdocView.swift
```diff
@@ -76,6 +76,8 @@ public struct ShareMdocQR: View {
                 authorizeBluetooth
             case .connecting(let qrPayload):
                 connectingView(qrPayload: qrPayload)
+            case .connectingViaNfc:
+                connectingViaNfcView
```
Juliano1612: I believe this state will never be reached, as this is the QR-engagement holder view, right? Worth adding a comment saying something like `ShareMdocView only uses QR engagement` to avoid someone trying to understand this call in the future.

ios/Showcase/Targets/AppUIKit/Sources/credentials/NfcHolderSelfTestView.swift
```diff
@@ -0,0 +1,292 @@
...
+                delegate: readerDelegate,
+                requestedItems: ["org.iso.18013.5.1": ["given_name": false, "family_name": false]],
+                trustAnchorRegistry: TrustedCertificatesDataStore.shared.getAllCertificates().map { $0.content },
+                l2capUsage: .disableL2CAP
+            )
+            append("Holder scanning + reader advertising over BLE…")
+
+            DispatchQueue.main.asyncAfter(deadline: .now() + 15) { [weak self] in
+                self?.fail("Holder/reader did not reach ready states in time — check Bluetooth permission")
+            }
```
Juliano1612, line 130: The 15s timeout is scheduled but never canceled. 

Run 1 succeeds at t=2s > re-run at t=5s > run 1's stale timeout still fires at t=15s and calls fail() on the second run (since finished was reset to false). 

You could store a `DispatchWorkItem` and `.cancel()` it in `stop()/run()`, or guard with a run-generation counter

Juliano1612 (approved):
Approving to avoid block merging, but left a few comments to check

ios/MobileSdk/Sources/MobileSdk/mdoc/proximity/PresentationManager.swift
tjzhang-spruce: Agreed. NFC static handover is central-client-mode only, so forcing central is intended — but I'll log a warning when no central mode is passed and document that on the .NFC case. Will update.

ios/MobileSdk/Sources/MobileSdk/mdoc/proximity/PresentationManager.swift
tjzhang-spruce: Yes will release as 0.18.x

ios/Showcase/Targets/AppUIKit/Sources/credentials/NfcHolderSelfTestView.swift
tjzhang-spruce: Good catch — real bug. The stale timer from a prior run can fire after a re-run resets finished. Will guard it with a run-generation token (and cancel on stop()). Fixing.

ios/Showcase/Targets/AppUIKit/Sources/credentials/ShareMdocView.swift
tjzhang-spruce: Good point — ShareMdocView is QR-only, so .connectingViaNfc never fires here; the branch only exists to satisfy the exhaustive switch. I'll drop the dead "hold near reader" view and make it EmptyView() with a comment noting this view is QR-only.

PR #339, opened 2026-07-06
ACC-1374 Adopt per credential key V1

leoliu-spruce:
## Description

This PR fixes a privacy limitation where OID4VP presentations previously signed all credentials with a shared holder key, which could enable correlation across sessions. This PR migrates to per-credential signing keys to prevent that.

**Approach: the credential carries its own key.** Each credential persists its own `key_alias`, which is the same field mdoc already uses. At presentation the holder resolves each credential's signing key from `ParsedCredential.key_alias()`, falling back to a shared `key_id` for legacy credentials that have no alias. So existing credentials keep working with no migration or re-issuance.

To make a per-credential key actually get stored, the credential-add path takes a key alias (see below).

## API changes

**Presentation key resolution (Rust core):**
- The OID4VP v1 and Draft 18 holders resolve each credential's key and thread it through to signing.

**Credential storage**
- `CredentialPack.tryAddAnyFormat(rawCredential, mdocKeyAlias)` is replaced by `tryAddAnyFormatWithKey(rawCredential, keyAlias)` (Kotlin/Swift). The pigeon `addAnyFormat` is likewise replaced by `addAnyFormatWithKey`.
  - Non-null `keyAlias` binds that key to the credential across **all** formats (SD-JWT VC, JWT-VC, mdoc), so its `key_alias` is the key used to sign its presentation.
  - Null stores keyless (`key_alias == None`) and presents via the fallback `keyId` — valid for issuer-signed non-mdoc formats. mdoc is device-bound and has no keyless form, so a null alias for an mdoc payload is rejected rather than binding a mismatched key.
  -  At presentation, the key must match the `cnf` fixed at issuance

### Migration (preserve existing single-key behavior)

- **Implementing a signer:** add the `keyId` parameter to the four methods. A single-key signer can ignore it and keep using its existing key.
- **Constructing a holder:** pass your existing key id as `keyId`. Every credential with no `key_alias` then signs with that one key, exactly as before.
- **Adding a credential:** move to `tryAddAnyFormatWithKey` / `addAnyFormatWithKey` and pass the key matching the credential's holder binding (`cnf` for non-mdoc, the MSO device key for mdoc), or null to store keyless and present via the fallback key. Passing the old mdoc-only alias unchanged is the one migration that compiles but is wrong for non-mdoc formats.

## Interactions with `main`

- **VCALM moved to the `vcalm-rs` crate.** The per-credential key handling that lived in `rust/src/vcalm/holder.rs` now has no home in this repo; `VcalmSignerAdapter` pins the session's `key_id` instead.
- **Multi-credential OID4VCI offers.** The showcases now issue every credential in an offer. Each offer generates one fresh `credential/<uuid>` key and binds the whole batch to it, because that key also backs the session's `client_id` and therefore every proof in the offer. Per-credential keys within a single offer would need a separate client id per credential.

## Known gaps

These are worth deciding on before merge rather than discovering later.

- **VCALM no longer resolves per-credential keys.** `vcalm-rs`'s signer port takes no `key_id`, so a VCALM session signs with the shared session key and ignores per-credential aliases. This also means the `MixedCredentialKeys` fail-fast guard added in review (for the case where a multi-credential VP mixes distinct holder keys and only the first credential's binding verifies) did not survive the extraction. Closing this needs a change in `vcalm-rs`.


SeaSkyThe (commented):
I tested on iOS (+ Animo Playground) and everything works great to me!


I'll leave without explicit approval and wait for Juliano's second look (since we have breaking changes for this one).

PS: Maybe update the Flutter example app to use the new structure

android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/CredentialPack.kt
```diff
@@ -115,21 +157,47 @@ class CredentialPack {
...
+                    message = "Could not parse credential without a keyAlias. mdoc credentials " +
+                        "require the device-key alias provisioned at issuance. Credential = $rawCredential",
+                    cause = e
+                )
+            }
+        }
+        val keyManager = KeyManager()
+        if (!keyManager.keyExists(keyAlias)) {
+            keyManager.generateSigningKey(keyAlias)
+        }
```
Juliano1612, line 230: This silently mints a fresh key for an unknown alias, but the Flutter signers explicitly refuse to do that ("a per-credential key must already exist from issuance, else it can't match the cnf binding"). A typo'd alias here produces a credential that stores fine but can never present successfully. Consider throwing instead, or documenting why add-time generation is safe.

android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/CredentialPack.kt
```diff
@@ -115,21 +157,47 @@ class CredentialPack {
...
      *
      * @param rawCredential The raw credential data as a string
-     * @param mdocKeyAlias The key alias to use if parsing as mdoc is needed
+     * @param keyAlias The per-credential key alias, or null to store keyless
      * @return List of parsed credentials
      * @throws ParsingException if the credential cannot be parsed in any supported format
      */
     @Throws(ParsingException::class)
-    fun tryAddAnyFormat(rawCredential: String, mdocKeyAlias: String): List<ParsedCredential> {
+    fun tryAddAnyFormat(rawCredential: String, keyAlias: String? = null): List<ParsedCredential> {
```
Juliano1612: This has a silent behavior change. This parameter used to only apply to mdoc, now it binds every format to that key. Credentials already stored (with `key_alias == None`) are unaffected, as they present via the holder's fallback key. But existing Kotlin/Dart call sites passing a dedicated mdoc alias still compile unchanged, and will store new SD-JWTs/JWT-VCs with a key_alias that doesn't match their cnf (verifiers reject those presentations, and we have no changelog to warn anyone). The break needs to be loud in code: either rename the method (keeping a deprecated mdocKeyAlias overload with the old behavior) so old callers fail to compile, or validate the alias's public key against the credential's cnf at add time and throw on mismatch.

Juliano1612: same for iOS and Flutter

android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/CredentialPack.kt
Juliano1612: same for iOS

rust/src/vcalm/holder.rs
```diff
@@ -991,8 +995,18 @@ impl VcalmHolder {
...
+        // credential's own key; a multi-credential VP uses the primary
+        // (first-presented) credential's key, falling back to the shared key when
+        // the credential has no alias (e.g. legacy credentials).
+        let active_key_id = present
+            .first()
+            .and_then(|cred| cred.key_alias())
+            .map(|alias| alias.0)
+            .unwrap_or_else(|| self.key_id.clone());
+        let holder_did = self.signer.did(active_key_id.clone());
+        let glue = VpSigner::new(self.signer.clone(), active_key_id, self.context_map.clone());
```
Juliano1612: In a multi-credential VP where credentials carry different `key_aliases`, only the first credential's holder binding will match the VP signature, and the others will fail verification. If mixed-key VPs aren't supported, we need to 1) fail fast with an explicit error rather than producing a partially verified VP or 2) implement it to support mixed-key VPs.

ios/Showcase/Targets/AppUIKit/Sources/wallet/HandleOID4VPView.swift
```diff
@@ -44,21 +45,22 @@ class Signer: PresentationSigner {
...
-    func did() -> String {
-        return try! didJwk.didFromJwk(jwk: _jwk)
+    func did(keyId: String) -> String {
+        return try! didJwk.didFromJwk(jwk: jwkFor(keyId))
     }
 
-    func jwk() -> String {
-        return _jwk
+    func jwk(keyId: String) -> String {
+        return try! jwkFor(keyId)
```
Juliano1612: A missing key (e.g. credential restored from backup on a new device, or deleted keystore entry) is a `try!` crash mid-presentation.

ios/Showcase/Targets/AppUIKit/Sources/wallet/Draft18OID4VPSupport.swift
```diff
@@ -30,24 +35,25 @@ class Draft18Signer: Draft18PresentationSigner {
...
     }
 
     func cryptosuite() -> String {
         "ecdsa-rdfc-2019"
     }
 
-    func jwk() -> String {
-        jwkString
+    func jwk(keyId: String) -> String {
+        return try! jwkFor(keyId)
```
Juliano1612: A missing key (e.g. credential restored from backup on a new device, or deleted keystore entry) is a `try!` crash mid-presentation.

android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/CredentialPack.kt
```diff
@@ -115,21 +157,47 @@ class CredentialPack {
...
+        if (keyAlias == null) {
+            return try {
+                tryAddRawCredential(rawCredential)
+            } catch (e: Exception) {
+                throw ParsingException(
+                    message = "Could not parse credential without a keyAlias. mdoc credentials " +
+                        "require the device-key alias provisioned at issuance. Credential = $rawCredential",
```
Juliano1612: Any parse failure on the null-alias path gets this mdoc-specific message, even for plain garbage input. We should keep the generic "unsupported format" message and add the mdoc hint as a secondary note.

Juliano1612: same for iOS

rust/src/oid4vp/facade.rs
```diff
@@ -177,12 +178,12 @@ pub struct Oid4vpPermissionResponse {
 #[uniffi::export(callback_interface)]
 #[async_trait::async_trait]
 pub trait Oid4vpPresentationSigner: Send + Sync + std::fmt::Debug {
-    async fn sign(&self, payload: Vec<u8>) -> Result<Vec<u8>, Oid4vpFacadeError>;
+    async fn sign(&self, key_id: String, payload: Vec<u8>) -> Result<Vec<u8>, Oid4vpFacadeError>;
     fn algorithm(&self) -> Algorithm;
```
Juliano1612, line 189: The `key_id` threading is incomplete as `algorithm()`/`cryptosuite()` don't have the `key_id`parameter. Artifacts signed with a per-credential key get their `alg`/`suite` from a different key (the fallback) or a hardcoded value. Harmless while every key is P-256/ES256, but nothing enforces that.

The same problem exists in all three signer traits: `oid4vp/presentation.rs`, `oid4vp/draft18/presentation.rs`, `oid4vp/facade.rs`. Suggest documenting the "all keys share one algorithm" assumption on the traits, since keying `algorithm()` would touch all of these.

rust/src/oid4vp/permission_request.rs
```diff
@@ -328,18 +333,28 @@ impl PermissionRequest {
...
         let mut vp_token_map: HashMap<String, Vec<VpTokenItem>> = HashMap::new();
 
         for cred in &selected_credentials {
+            let key_alias = ParsedCredential {
+                inner: cred.inner.clone(),
+            }
+            .key_alias();
+            let key_id = key_alias
+                .map(|alias| alias.0)
+                .unwrap_or_else(|| self.key_id.clone());
```
Juliano1612: Building a throwaway `ParsedCredential` just to read `key_alias` is a smell, and the same resolve-or-fallback snippet appears in both `permission_request` modules (and a third variant in `vcalm/holder.rs`). Expose `key_alias()` on `PresentableCredential`/`Draft18PresentableCredential` and add a small `resolve_key_id(fallback)` helper.

android/Showcase/src/main/java/com/spruceid/mobilesdkexample/wallet/HandleOID4VCIView.kt
```diff
@@ -52,6 +52,9 @@ fun HandleOID4VCIView(
     var loading by remember { mutableStateOf(false) }
     var err by remember { mutableStateOf<String?>(null) }
     var credential by remember { mutableStateOf<String?>(null) }
+    // The fresh per-credential key alias generated for this issuance, carried
+    // to AddToWalletView so it is stored as the credential's key_alias.
```
Juliano1612:
```suggestion
```

ios/Showcase/Targets/AppUIKit/Sources/wallet/Draft18OID4VPSupport.swift
```diff
@@ -3,25 +3,30 @@ import SpruceIDMobileSdkRs
 import SwiftUI
 
 class Draft18Signer: Draft18PresentationSigner {
+    // The shared/legacy fallback key. Per-call methods resolve the actual key
+    // from the `keyId` Rust passes; this is only used where no per-call key
+    // applies (e.g. `algorithm()`).
```
Juliano1612:
```suggestion
    // The shared/legacy fallback key.
```

Juliano1612 (changes requested):

android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/CredentialPack.kt
leoliu-spruce: I believe this mistakenly mirrored the mdoc path. Throwing here now

rust/src/vcalm/holder.rs
leoliu-spruce: Going with the fail fast approach and erroring on the first mistmatch

rust/src/oid4vp/permission_request.rs
leoliu-spruce: Added `resolve_key_id` here

rust/src/oid4vp/facade.rs
leoliu-spruce: Added documentation explaining all the keys share the same algorithm

android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/CredentialPack.kt
leoliu-spruce: Changed the message to "The credential format is not supported"

ios/Showcase/Targets/AppUIKit/Sources/wallet/Draft18OID4VPSupport.swift
leoliu-spruce: Returning an empty string on missing key.

android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/CredentialPack.kt
leoliu-spruce: renamed the methods

Ryanmtate (approved):

Juliano1612 (approved):
Approving, but one thing worth a follow-up: on a fresh install, adding a credential via the `spruceid://?sd-jwt=` deep link now fails silently. `acceptRawCredentialIntoWallet` falls back to `DEFAULT_SIGNING_KEY_ID`, which nothing has created yet at that point (it's only minted lazily by the OID4VP signers / mock mDL), so `tryAddAnyFormatWithKey` throws and `AddToWalletView` swallows it as a decline, and the user just sees "0 of 1 credentials accepted". It's Showcase (both platforms) only, but it used to work before the "key must exist" check. Easiest fix is probably bootstrapping the default key in the fallback path, same as the `Signer` init does.

PR #340, opened 2026-07-06
Add verify_device_response function to for mdl reader

theosirian:
## Description

Adds `verify_device_response` for Apple ID Verify RawData requests, the iOS framework returns `device_response`, `session_transcript` and `ephemeral_reader_key` (>=26.4), and the current code only handles verification and decoding of a `device_response` when it's negotiated and obtained through the `MDLSessionManager`.

### DIFF
--- a/rust/src/mdl/reader.rs
+++ b/rust/src/mdl/reader.rs
@@ -218,19 +218,10 @@ pub fn establish_session(
                 value: format!("Unable to build namespaces: {e:?}"),
             })?;
 
-    let registry = TrustAnchorRegistry::from_pem_certificates(
-        trust_anchor_registry
-            .into_iter()
-            .flat_map(|v| v.into_iter())
-            .map(|certificate_pem| PemTrustAnchor {
-                certificate_pem,
-                purpose: x509::trust_anchor::TrustPurpose::Iaca,
-            })
-            .collect(),
-    )
-    .map_err(|e| MDLReaderSessionError::Generic {
-        value: format!("unable to construct TrustAnchorRegistry: {e:?}"),
-    })?;
+    let registry =
+        build_registry(trust_anchor_registry).map_err(|e| MDLReaderSessionError::Generic {
+            value: format!("unable to construct TrustAnchorRegistry: {e:?}"),
+        })?;
 
     let (manager, request, ble_ident) =
         reader::SessionManager::establish_session(handover.0.clone(), namespaces, registry)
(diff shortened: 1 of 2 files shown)


rust/src/mdl/reader.rs
```diff
@@ -427,3 +444,160 @@ pub fn handle_response(
...
+        }
+    })
+}
+
+#[uniffi::export]
+pub fn verify_device_response(
+    device_response: Vec<u8>,
+    session_transcript: Vec<u8>,
+    ephemeral_reader_key: Vec<u8>,
+    trust_anchor_registry: Option<Vec<String>>,
```
Juliano1612, line 515: can you add doc comments specifying the expected encoding of each parameter?

rust/src/mdl/reader.rs
```diff
@@ -427,3 +444,160 @@ pub fn handle_response(
...
+            namespaces,
+            doc_types,
+            &(),
+            if ephemeral_reader_key.is_empty() {
+                [0u8; 32]
+            } else {
+                ephemeral_reader_key
+                    .try_into()
+                    .map_err(|e| MDLReaderResponseError::Generic {
+                        value: format!("unable to parse ephemeral_reader_key: {e:?}"),
```
Juliano1612: is this printing the rejected `Vec<u8>`? if yes, we are printing the raw private key bytes

rust/src/mdl/reader.rs
```diff
@@ -427,3 +444,160 @@ pub fn handle_response(
...
+    let registry = TrustAnchorRegistry::from_pem_certificates(
+        trust_anchor_registry
+            .into_iter()
+            .flatten()
+            .map(|certificate_pem| PemTrustAnchor {
+                certificate_pem,
+                purpose: x509::trust_anchor::TrustPurpose::Iaca,
+            })
+            .collect(),
+    )
```
Juliano1612: This block is a copy of the registry construction in `establish_session` (reader.rs:221-233), differing only in the error type. Extract a small helper returning `Result<TrustAnchorRegistry, String>` and map the error at each call site

rust/src/mdl/reader.rs
```diff
@@ -427,3 +444,160 @@ pub fn handle_response(
...
+    let (document, x5chain, namespaces) =
+        reader::parse(&device_response).map_err(|e| MDLReaderResponseError::Generic {
+            value: format!("unable to parse device response: {e:?}"),
+        })?;
+
+    let doc_types: Vec<String> = device_response
+        .documents
+        .as_ref()
+        .map(|docs| docs.iter().map(|d| d.doc_type.clone()).collect())
+        .unwrap_or_default();
```
Juliano1612, line 545: Claude finding: 
`reader::parse` (in isomdl, untouched here) returns only the first document of the `DeviceResponse`, so issuer/device authentication runs against that one document — while `doc_types` reports every document in the response. A multi-document response would show all doc types as if verified. This mirrors the existing `handle_response` behavior, but since this API accepts arbitrary external responses, add a doc note (or an error/warning when `documents.len() > 1`).

Juliano1612 (approved):
LGTM! just a few small comments to look and fix

PR #341, opened 2026-07-07
ACC-1410: Support multi-credential OID4VCI offers in iOS and Android Showcase

kshaunishaddala-cpu:
Support multi-credential OID4VCI offers in iOS and Android Showcase apps

## Description

An OID4VCI offer can list more than one `credential_configuration_id`. Before this PR, both apps called `defaultCredentialId()` to pick a single credential to exchange — which errors out (`AmbiguousCredentialOffer`) whenever an offer has more than one configuration ID. So multi-credential offers failed entirely instead of being handled.

This PR makes `HandleOID4VCIView` loop over every `credential_configuration_id` in the offer, exchanging each with its own fresh nonce/proof, and makes `AddToWalletView` accept a list of credentials instead of one — rendering a review card per credential with a single Add/Decline action for the batch. Same pattern on iOS and Android. The Rust/FFI layer and Flutter plugin already supported this; the gap was app-layer only.

_Summarize the changes included in this pull request. What is the current behavior and what is the updated/expected behavior of this PR? Include relevant context, and reasoning behind architectural/design decisions._

### Design / Video

https://github.com/user-attachments/assets/10f431f4-883d-45ed-975f-8e6d5b3e9c1a

https://github.com/user-attachments/assets/18702a87-bd9c-48cf-a69b-62f8614833af

### Other changes

_Describe any minor or "drive-by" changes here._

### Optional section

* Reviewers, please pay special attention to...
* If this PR exceeds 500 lines, please explain why

## Tested

Ran the `oid4vci-rs` example server with a 3-credential config (`cargo run --example server -- examples/server-multi.json --pre-auth`), exposed via `cloudflared`, and drove the full flow on a physical iOS and Android device via deep link. Confirmed all 3 credentials were exchanged and saved. Single-credential paths and the existing deep-link entry points verified by inspection (list-of-one behaves identically to the old single value).

### DIFF
--- a/android/Showcase/src/main/java/com/spruceid/mobilesdkexample/credentials/AddToWalletView.kt
+++ b/android/Showcase/src/main/java/com/spruceid/mobilesdkexample/credentials/AddToWalletView.kt
@@ -178,7 +178,9 @@ fun AddToWalletView(
             Box(Modifier.weight(1f)) {
                 HorizontalPager(
                     state = pagerState,
-                    userScrollEnabled = false,
+                    // Swiping back to an already-decided step is safe: the
+                    // button bar below only renders for undecided, parsed
+                    // steps, guarded by `decidedIndices`.
                     modifier = Modifier.fillMaxSize()
                 ) { page ->
                     when (val step = stepItems[page]) {
--- a/android/Showcase/src/main/java/com/spruceid/mobilesdkexample/credentials/AddToWalletView.kt
+++ b/android/Showcase/src/main/java/com/spruceid/mobilesdkexample/credentials/AddToWalletView.kt
@@ -4,10 +4,13 @@ import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.navigationBarsPadding
 import androidx.compose.foundation.layout.padding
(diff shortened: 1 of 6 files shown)


SeaSkyThe (changes requested):
Would you mind changing the credential acceptance flow to a step-by-step experience?

Instead of displaying all the credentials on a single page, could you iterate through them and show the same screen we had before for each credential? The only difference would be adding a stepper at the top of the screen to indicate the user's progress. And rejecting or accepting each credentials should be independent actions. At the end you can show a toast (if we have it configured, otherwise don't bother) indicating that "X of Y credentials accepted".


Can you also add videos showing the flow for both android and ios?

SeaSkyThe (commented):
Heyy. Cool stuff, looking good! I have a question, do you know if there is any particular reason the details of the credential being accepted is not showing in the Android version? (It can be just a UI problem of things being hidden by the layout)

Small UI changes/nitpicks:
1. In Android, can you align the content at the top, like in iOS? So the "Review Info" is always close to the stepper? 
    - And also, make the Add to Wallet and Cancel buttons be always at the bottom? Same as in iOS
2. Can you add a little padding between the stepper and the "Review Info" header? 
3. Maybe the stepper in the Android version is not respecting the Safe area, meaning its rendering below the system information at the top of the screen?

SeaSkyThe (approved):
LGTM! Tested both regression (single credential offers) and the new flow for multi-credential and both work!

Not a blocker: One little UX thing that I liked before the [allowing acceptance of creds in any order](https://github.com/spruceid/sprucekit-mobile/pull/341/commits/3ffb56ec82d38739bac625baee764543a0b47fc4) commit was the old `advance()` helper (which did `animateScrollToPage(currentIndex + 1)`) called after we accept a credential, which made page go to the next credential to be accepted. The new approach replaced it with `finishIfAllDecided()`, which only navigates home once all pages are decided, so it never scrolls the pager (the user has to do it manually) which can be confusing to the user.

PR #342, opened 2026-07-07
Explicitly generate a random 96-bit IV from OS entropy (CWE-327 defense-in-depth)

patmalt:
**Finding:** `javax.crypto.spec.GCMParameterSpec.!operator_javanewinit` ([CWE-327](https://cwe.mitre.org/data/definitions/327.html))
**Module:** `com.spruceid.mobile.sdk.KeyManager` (Android SDK)

### Finding summary

Flagged `GCMParameterSpec(128, iv)` constructor calls in `KeyManager.kt`, asserting the initialization vector is "not cryptographically strong". However, this assertion was a false positive.

### Why this is a false positive

**IVs are generated from OS entropy at encryption time.** Encryption paths initialize the cipher without an explicit IV, so the IV is generated by the crypto provider:
   - `encryptPayload` uses AndroidKeyStore-backed keys (StrongBox/TEE). The Keystore generates a random 96-bit IV inside secure hardware. AndroidKeyStore *rejects* caller-supplied IVs on encrypt unless `setRandomizedEncryptionRequired(false)` is set — this code does not opt out.
   - `encryptWithDirectKey` used Conscrypt's implicit provider-generated random IV.

### Hardening applied (defense-in-depth)

The following changes were made in `KeyManager.kt`:

- `encryptWithDirectKey` now explicitly generates a 96-bit IV via `java.security.SecureRandom` (seeded from OS entropy) rather than relying on provider defaults.
- Magic numbers replaced with named constants (`GCM_IV_LENGTH = 12`, `GCM_TAG_LENGTH_BITS = 128`).

## References

- [CWE-327](https://cwe.mitre.org/data/definitions/327.html)
- [Android Keystore: `setRandomizedEncryptionRequired`](https://developer.android.com/reference/android/security/keystore/KeyGenParameterSpec.Builder#setRandomizedEncryptionRequired(boolean))

### DIFF
--- a/android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/KeyManager.kt
+++ b/android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/KeyManager.kt
@@ -9,14 +9,13 @@ import androidx.annotation.RequiresApi
 import com.spruceid.mobile.sdk.rs.CryptoCurveUtils
 import com.spruceid.mobile.sdk.rs.Jwk
 import com.spruceid.mobile.sdk.rs.KeyAlias
-import com.spruceid.mobile.sdk.rs.KeyStore as SpruceKitKeyStore
 import com.spruceid.mobile.sdk.rs.SigningKey
 import com.spruceid.mobile.sdk.rs.coseKeyEc2P256PublicKey
 import com.spruceid.mobile.sdk.rs.jwkFromPublicP256
 import java.security.KeyPairGenerator
 import java.security.KeyStore
+import java.security.SecureRandom
 import java.security.Signature
-import java.security.cert.Certificate
 import java.security.interfaces.ECPublicKey
 import java.security.spec.ECGenParameterSpec
 import javax.crypto.Cipher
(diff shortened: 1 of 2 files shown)


SeaSkyThe (approved):

android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/KeyManager.kt
```diff
@@ -381,8 +389,9 @@ class KeyManager : SpruceKitKeyStore {
         val cipher = Cipher.getInstance("AES/GCM/NoPadding")
         val secretKey = SecretKeySpec(key, "AES")
 
-        cipher.init(Cipher.ENCRYPT_MODE, secretKey)
-        val iv = cipher.iv
+        val iv = ByteArray(GCM_IV_LENGTH).also { SecureRandom().nextBytes(it) }
```
Juliano1612: Minor: `SecureRandom()` is created on every encrypt call. Can we move it to the companion object?

Juliano1612 (approved):

android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/KeyManager.kt
patmalt: Updated

PR #346, opened 2026-07-13
Extract StorageManager, HttpClient, KeyStore into separate crate

w4ll3:
## Description

This PR moves a few SpruceKit types to an external crate and adds it as a dependency, enabling reuse of those types in other dependencies and ensuring type consistency. It is not supposed to change any functionality or feature.

### Other changes

Had to bridge a few types internally.

## Tested

Tested
- Credible Wallet (iOS/Android)
- Safe (iOS/Android)
- SpruceKit Mobile (iOS/Android)


android/MobileSdk/build.gradle.kts
```diff
@@ -43,12 +43,12 @@ publishing {
...
-        //    artifactId = "mobilesdk"
-        //    version = System.getenv("VERSION")
-        //  afterEvaluate { from(components["release"]) }
-        // }
+        create<MavenPublication>("debug") {
+           groupId = "com.spruceid.mobile.sdk"
+           artifactId = "mobilesdk"
+           version = System.getenv("VERSION")
+         afterEvaluate { from(components["release"]) }
+        }
```
Juliano1612, line 53: comment the debug publication

flutter/example/android/gradle.properties
```diff
@@ -1,3 +1,5 @@
 org.gradle.jvmargs=-Xmx8G -XX:MaxMetaspaceSize=4G -XX:ReservedCodeCacheSize=512m -XX:+HeapDumpOnOutOfMemoryError
 android.useAndroidX=true
 android.enableJetifier=true
+android.builtInKotlin=false
+android.newDsl=false
```
Juliano1612: is this necessary?

rust/src/oid4vp/dynamic_credential.rs
```diff
@@ -0,0 +1,127 @@
...
+pub trait DynamicCredentialProvider: Send + Sync {
+    /// Credentials this provider could mint to satisfy `query`.
+    ///
+    /// Surfaced in the [`crate::oid4vp::PermissionRequest`] so the UI can let
+    /// the user opt in (or decline). An empty vector means this provider cannot
+    /// satisfy the query.
+    ///
+    /// `query` is the DCQL credential query as JSON; use
+    /// [`DcqlCredentialQueryJson::parse`] to inspect it.
+    fn offers(&self, query: DcqlCredentialQueryJson) -> Vec<DynamicCredentialOffer>;
```
Juliano1612: @Ryanmtate `offers` is a sync foreign call and the holder invokes it inline (provider x query loop) while building the permission request, so a slow implementation on the Kotlin/Swift side blocks the whole flow. Since `mint` is already async, maybe make this async too for symmetry? Cheaper now than as a breaking change later. If we keep it sync, worth a doc note saying it must be fast/local and heavy work belongs in `mint`

android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/KeyManager.kt
```diff
@@ -8,8 +8,9 @@ import android.util.Log
 import androidx.annotation.RequiresApi
 import com.spruceid.mobile.sdk.rs.CryptoCurveUtils
 import com.spruceid.mobile.sdk.rs.Jwk
-import com.spruceid.mobile.sdk.rs.KeyAlias
-import com.spruceid.mobile.sdk.rs.SigningKey
+import uniffi.mobile_toolkit.KeyAlias
+import uniffi.mobile_toolkit.KeyStore as SpruceKitKeyStore
+import uniffi.mobile_toolkit.SigningKey
```
Juliano1612: this is a breaking change for Kotlin consumers. Also, `uniffi.mobile_toolkit` is just the uniffi default package name. If this crate is meant to be reused by other SDKs, might be worth setting a proper package_name in `mobile-toolkit/uniffi.toml` now (e.g. `com.spruceid.toolkit`), otherwise renaming it later will be a second breaking change just like this one.

Or is there a way to re-export it without releasing it as a breaking change?

rust/mobile-toolkit/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "mobile-toolkit"
+version = "0.15.2"
```
Juliano1612: can you check what is better here? are we releasing this with all other packages in the monorepo? Will we keep the same global version here? can you think a little bit more about the release process?

android/MobileSdkRs/build.gradle
```diff
@@ -77,14 +77,30 @@ def markerFile(taskName) {
...
     if (!markerFile.exists()) {
         return true
     }
 
+    if (taskName.startsWith("buildCargoNdk")
+            && fileTree(dir: "src/main/jniLibs", include: "**/*.so").isEmpty()) {
+        return true
+    }
+    if (taskName == "bindGen"
+            && fileTree(dir: "src/main/java", include: "**/mobile_sdk_rs.kt").isEmpty()) {
```
Juliano1612:
```suggestion
            && (fileTree(dir: "src/main/java", include: "**/mobile_sdk_rs.kt").isEmpty()
                    || fileTree(dir: "src/main/java", include: "**/mobile_toolkit.kt").isEmpty())) {
```

rust/Cargo.toml
```diff
@@ -1,6 +1,10 @@
+[workspace]
```
Juliano1612, line 1: Found a blocker while testing this branch on a physical Android device (arm64): the app crashes at startup with UniFFI API checksum mismatch in uniffi.mobile_toolkit, before anything else runs.

It's not this PR's code, it's the uniffi bump from 0.31.0 to 0.31.2 in Cargo.lock (came in with the "update rust bindings" commit). 0.31.2 has a known regression on AArch64: checksum values >= 0x8000 come back sign-extended from the native side and the Kotlin bindings now compare them as a 32-bit Int, so validation always fails on device. Upstream report: mozilla/uniffi-rs#2939, pending fix: mozilla/uniffi-rs#2935. It doesn't reproduce on mac or CI (Apple ABI zero-extends the register), which is probably why it slipped through.

Confirmed both ways locally: disassembled the checksum functions in the packaged .so (values are correct as u16, just sign-extended in the register), and after pinning uniffi to 0.31.1 and regenerating the bindings the app starts fine on the same device.

Suggested fix until upstream ships the patch: pin uniffi = "=0.31.1" in both rust/Cargo.toml and rust/mobile-toolkit/Cargo.toml and regenerate the bindings, including the committed Swift ones so the CI diff check stays green.

Juliano1612 (changes requested):

rust/Cargo.toml
w4ll3: Pinned =0.31.1 via workspace.dependencies with a comment pointing at mozilla/uniffi-rs#2939

rust/mobile-toolkit/Cargo.toml
w4ll3: mobile-toolkit isn't released on its own it's statically included in the mobile-sdk-rs artifacts so it keeps the global version, now enforced via workspace version inheritance.

flutter/example/android/gradle.properties
w4ll3: removed

android/MobileSdk/src/main/java/com/spruceid/mobile/sdk/KeyManager.kt
w4ll3: bindings now generate into com.spruceid.mobile.toolkit, and ToolkitTypes.kt re-exports the moved types from com.spruceid.mobile.sdk.rs via typealiases, so existing app imports keep compiling

Juliano1612 (approved):
LGTM! tested on Android and iOS

PR #349, opened 2026-07-29
Protocol discovery function (for interaction URLs)

aashishah-eng:
## Description

**Added:** `discover_protocols`, to fetch all protocols for an interaction URL
**Currently** `VcalmHolder.start_exchange` discovers all protocols, but only uses the `vcapi` key, throwing the other protocol data away. As a result, implementations in Swift/Kotlin have to make a separate GET call to discover protocols, offer them ALL to the user, then leverage functions like `start_exchange` only if the user selected `VCALM`. 
**Updated behaviour** allows the Rust side to continue discovering all the protocols.  `discover_protocols` is implemented as a free function (not tied to a struct) to make it accessible regardless of the protocol

### Other changes
Wired in the use of `discover_protocols` into `VcalmHolder.start_exchange`, and passed appropriate error types to `VcalmError`

## Tested
Existing test suites in `rust/src/vcalm/holder.rs` were validated, and use of `discover_protocols` in the SwiftUI layer was validated (is accessible and functions as expected)

### DIFF
--- a/rust/src/discover_protocols.rs
+++ b/rust/src/discover_protocols.rs
@@ -43,7 +43,7 @@ pub enum DiscoveryError {
 pub(crate) const MAX_RESPONSE_BYTES: usize = 10 * 1024 * 1024;
 
 #[uniffi::export(async_runtime = "tokio")]
-// Standalone FFI entry point: no caller-supplied client, so build a fresh one
+/// Standalone FFI entry point: no caller-supplied client, so build a fresh one
 pub async fn discover_protocols(
     interaction_url: &str,
 ) -> Result<HashMap<String, String>, DiscoveryError> {
@@ -307,7 +307,6 @@ mod tests {
         let err = discover_protocols("file:///etc/passwd")
             .await
             .expect_err("file: scheme must be rejected");
-        println!("{}", err);
         assert!(matches!(err, DiscoveryError::InsecureUrl(_)));
     }
 
--- a/rust/src/discover_protocols.rs
+++ b/rust/src/discover_protocols.rs
@@ -1,0 +1,142 @@
+use reqwest::header::ACCEPT;
+use std::collections::HashMap;
+use std::time::Duration;
+use url::Url;
+
+/// The discovery document returned for an `interaction:` initiation (§3.7.4).
+#[derive(serde::Deserialize)]
+struct DiscoveryResponse {
+    protocols: HashMap<String, String>,
+}
+#[derive(thiserror::Error, Debug, uniffi::Error)]
+pub enum DiscoveryError {
+    /// A transport-level failure (stringified `reqwest::Error`).
+    #[error("Network error: {0}")]
+    Network(String),
+
+    // Interaction URL provided is invalid
+    #[error("invalid URL: {0}")]
+    InvalidUrl(String),
+
+    /// A 5xx (or otherwise non-2xx/4xx) server response.
+    #[error("Server returned error status {status}")]
+    ServerError { status: u16, body: String },
+
+    /// A response body failed to deserialize (stringified `serde_json::Error`).
+    #[error("Failed to deserialize response: {0}")]
+    Deserialization(String),
+
+    /// A response body exceeded the configured size cap (B.4).
+    #[error("response body exceeded the {limit_bytes}-byte limit")]
+    ResponseTooLarge { limit_bytes: u64 },
+
+    /// A non-HTTPS (or non-HTTP-scheme) URL was rejected (§3.7.1 / B.2). Plain
+    /// `http` is only accepted for loopback hosts (local development).
+    #[error("insecure URL rejected: {0}")]
+    InsecureUrl(String),
+}
+
+/// Cap on a discovery/exchange response body (B.4: large payloads can trigger
+/// DoS incidents — a malicious or broken server must not be able to exhaust the
+/// wallet's memory). Generous for any plausible VPR/VP payload.
+pub(crate) const MAX_RESPONSE_BYTES: usize = 10 * 1024 * 1024;
+
+#[uniffi::export(async_runtime = "tokio")]
+/// Resolve all protocol exchange URLs from an `interaction:` discovery endpoint.
+/// The discovery URL must pass [`validate_endpoint_url`] (HTTPS, or loopback http
+/// for local dev — §3.7.1/B.2; also rejects `file:`/other schemes a QR code could smuggle in).
+pub async fn discover_protocols(
+    discovery_url: &str,
+) -> Result<HashMap<String, String>, DiscoveryError> {
+    let discovery_url = Url::parse(discovery_url)
+        .map_err(|e| DiscoveryError::InvalidUrl(format!("invalid exchange URL: {e}")))?;
+    validate_endpoint_url(&discovery_url)?;
+
+    let client = reqwest::Client::builder()
+        .use_rustls_tls()
+        .timeout(Duration::from_secs(30))
+        .build()
+        .map_err(|e| DiscoveryError::Network(e.to_string()))?;
+
+    let resp = client
+        .get(discovery_url)
+        .header(ACCEPT, "application/json")
+        .send()
+        .await
+        .map_err(|e| DiscoveryError::Network(e.to_string()))?;
+
+    let status = resp.status();
+    let body = read_body_capped(resp).await?;
+
+    if !status.is_success() {
+        return Err(DiscoveryError::ServerError {
+            status: status.as_u16(),
+            body,
+        });
+    }
+
+    let discovery: DiscoveryResponse =
+        serde_json::from_str(&body).map_err(|e| DiscoveryError::Deserialization(e.to_string()))?;
+
+    Ok(discovery.protocols)
+}
+
+/// §3.7.1: The interaction URL must be an HTTPS URL that contains an interaction-specific identifier.
+/// The URL SHOULD be opaque and require no URL syntax processing before it is fetched by the receiving
+/// system — the HTTPS origin is the trust signal the whole interaction model hangs on, and a bearer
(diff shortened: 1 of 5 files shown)


rust/src/discover_protocols.rs
```diff
@@ -0,0 +1,142 @@
...
+#[uniffi::export(async_runtime = "tokio")]
+/// Resolve all protocol exchange URLs from an `interaction:` discovery endpoint.
+/// The discovery URL must pass [`validate_endpoint_url`] (HTTPS, or loopback http
+/// for local dev — §3.7.1/B.2; also rejects `file:`/other schemes a QR code could smuggle in).
+pub async fn discover_protocols(
+    discovery_url: &str,
+) -> Result<HashMap<String, String>, DiscoveryError> {
+    let discovery_url = Url::parse(discovery_url)
+        .map_err(|e| DiscoveryError::InvalidUrl(format!("invalid exchange URL: {e}")))?;
+    validate_endpoint_url(&discovery_url)?;
```
Juliano1612: Can you add the `iuv` version check here, right after parsing? The spec ([§3.7.1](https://www.w3.org/TR/vcalm-1.0/#interaction-url-format)) says `iuv` "MUST be 1 when using this version of this API", and this check currently only exists in `start_exchange`. Absent iuv stays allowed.
Add after `validate_endpoint_url`, something like:


```suggestion
    let discovery_url = Url::parse(discovery_url)
        .map_err(|e| DiscoveryError::InvalidUrl(format!("invalid exchange URL: {e}")))?;
    validate_endpoint_url(&discovery_url)?;
    if let Some((_, version)) = discovery_url.query_pairs().find(|(k, _)| k == "iuv") {
      if version != "1" {
          return Err(DiscoveryError::InvalidUrl(format!(
              "unsupported interaction URL version: iuv={v} (expected 1)"
          )));
      }
  }
```

rust/src/discover_protocols.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    let client = reqwest::Client::builder()
+        .use_rustls_tls()
```
Juliano1612:
```suggestion
        .use_rustls_tls()
        .redirect(reqwest::redirect::Policy::none())
```

rust/src/discover_protocols.rs
```diff
@@ -0,0 +1,142 @@
...
+
+#[uniffi::export(async_runtime = "tokio")]
+/// Resolve all protocol exchange URLs from an `interaction:` discovery endpoint.
+/// The discovery URL must pass [`validate_endpoint_url`] (HTTPS, or loopback http
+/// for local dev — §3.7.1/B.2; also rejects `file:`/other schemes a QR code could smuggle in).
+pub async fn discover_protocols(
+    discovery_url: &str,
+) -> Result<HashMap<String, String>, DiscoveryError> {
+    let discovery_url = Url::parse(discovery_url)
+        .map_err(|e| DiscoveryError::InvalidUrl(format!("invalid exchange URL: {e}")))?;
```
Juliano1612:
```suggestion
        .map_err(|e| DiscoveryError::InvalidUrl(format!("invalid interaction URL: {e}")))?;
```

rust/src/discover_protocols.rs
```diff
@@ -0,0 +1,142 @@
...
+pub(crate) const MAX_RESPONSE_BYTES: usize = 10 * 1024 * 1024;
+
+#[uniffi::export(async_runtime = "tokio")]
+/// Resolve all protocol exchange URLs from an `interaction:` discovery endpoint.
+/// The discovery URL must pass [`validate_endpoint_url`] (HTTPS, or loopback http
+/// for local dev — §3.7.1/B.2; also rejects `file:`/other schemes a QR code could smuggle in).
+pub async fn discover_protocols(
+    discovery_url: &str,
+) -> Result<HashMap<String, String>, DiscoveryError> {
+    let discovery_url = Url::parse(discovery_url)
```
Juliano1612: We should rename `discovery_url` to `interaction_url`

rust/src/discover_protocols.rs
```diff
@@ -0,0 +1,142 @@
...
+        if buf.len() + chunk.len() > MAX_RESPONSE_BYTES {
+            return Err(DiscoveryError::ResponseTooLarge {
+                limit_bytes: MAX_RESPONSE_BYTES as u64,
+            });
+        }
+        buf.extend_from_slice(&chunk);
+    }
+    String::from_utf8(buf)
+        .map_err(|e| DiscoveryError::Deserialization(format!("response body is not UTF-8: {e}")))
+}
```
Juliano1612, line 154: Please add a `#[cfg(test)]` module here with wiremock tests (see `vcalm/holder.rs` tests for the pattern). Minimum set:
- Happy path: server returns `{"protocols": {"vcapi": "...", "OID4VP": "openid4vp://..."}}` > full map returned, including the non-HTTP-scheme value (legit per spec Example 19)
- Assert the request carries `Accept: application/json` (the spec conditions the JSON response on it, §3.7.4)
- 500 response > `DiscoveryError::ServerError`
- Non-JSON body > `DiscoveryError::Deserialization`

rust/src/vcalm/error.rs
```diff
@@ -115,3 +115,20 @@ impl From<uniffi::UnexpectedUniFFICallbackError> for VcalmError {
...
+        match e {
+            DiscoveryError::ServerError { status, body } => {
+                VcalmError::ServerError { status, body }
+            }
+            DiscoveryError::Deserialization(v) => VcalmError::Deserialization(v),
+            DiscoveryError::ResponseTooLarge { limit_bytes } => {
+                VcalmError::ResponseTooLarge { limit_bytes }
+            }
+            DiscoveryError::InsecureUrl(v) => VcalmError::InsecureUrl(v),
+            other => VcalmError::Network(other.to_string()),
```
Juliano1612: With a catch-all, any future `DiscoveryError` variant silently maps to `Network` instead of producing a compile error reminding us to map it

```suggestion
            DiscoveryError::Network(v) => VcalmError::Network(v),
            DiscoveryError::InvalidUrl(v) => VcalmError::Network(v),
```

rust/src/discover_protocols.rs
```diff
@@ -0,0 +1,142 @@
...
+struct DiscoveryResponse {
+    protocols: HashMap<String, String>,
+}
+#[derive(thiserror::Error, Debug, uniffi::Error)]
+pub enum DiscoveryError {
+    /// A transport-level failure (stringified `reqwest::Error`).
+    #[error("Network error: {0}")]
+    Network(String),
+
+    // Interaction URL provided is invalid
```
Juliano1612: nit: change `//` to `///` so this shows up in rustdoc

```suggestion
    /// Interaction URL provided is invalid
```

rust/src/discover_protocols.rs
```diff
@@ -0,0 +1,142 @@
...
+/// The discovery URL must pass [`validate_endpoint_url`] (HTTPS, or loopback http
+/// for local dev — §3.7.1/B.2; also rejects `file:`/other schemes a QR code could smuggle in).
+pub async fn discover_protocols(
+    discovery_url: &str,
+) -> Result<HashMap<String, String>, DiscoveryError> {
+    let discovery_url = Url::parse(discovery_url)
+        .map_err(|e| DiscoveryError::InvalidUrl(format!("invalid exchange URL: {e}")))?;
+    validate_endpoint_url(&discovery_url)?;
+
+    let client = reqwest::Client::builder()
```
Juliano1612: this client config duplicates `VcalmHolder::new_session` (holder.rs:103-107), and they've already drifted with this Policy::none() suggestion I added. I suggest extracting a `pub(crate) fn build_http_client()` here and use it in both places. If changing the holder's redirect behavior feels out of scope, apply the helper here only and we'll unify the holder in a follow-up

Juliano1612 (changes requested):
Nice refactor! Just requesting changes for a few spec/hardening items before merge.

rust/src/discover_protocols.rs
aashishah-eng: I have both `holder` and `discover_protocols` using `build_http_client()` now. The holder can pass in its own client instance (`self.client`) to complete discovery on, and the uniffi entry point creates a one-off client for this purpose (discover_protocols:45-60)

rust/src/discover_protocols.rs
```diff
@@ -0,0 +1,332 @@
...
+        assert!(matches!(err, DiscoveryError::InsecureUrl(_)));
+    }
+
+    /// A `file:` URL (the sort a malicious QR code could smuggle in) is rejected.
+    #[tokio::test]
+    async fn file_scheme_is_insecure() {
+        let err = discover_protocols("file:///etc/passwd")
+            .await
+            .expect_err("file: scheme must be rejected");
+        println!("{}", err);
```
Juliano1612: nit
```suggestion
```

rust/src/discover_protocols.rs
```diff
@@ -0,0 +1,332 @@
...
+    InsecureUrl(String),
+}
+
+/// Cap on a discovery/exchange response body (B.4: large payloads can trigger
+/// DoS incidents — a malicious or broken server must not be able to exhaust the
+/// wallet's memory). Generous for any plausible VPR/VP payload.
+pub(crate) const MAX_RESPONSE_BYTES: usize = 10 * 1024 * 1024;
+
+#[uniffi::export(async_runtime = "tokio")]
+// Standalone FFI entry point: no caller-supplied client, so build a fresh one
```
Juliano1612: nit
```suggestion
/// Standalone FFI entry point: no caller-supplied client, so build a fresh one
```

Juliano1612 (commented):
LGTM!

Juliano1612 (approved):

PR #350, opened 2026-07-31
Showcase: update targetSdk from 35 to 36

SeaSkyThe:
## Description

Updated Sprucekit Showcase `targetSdk` from 35 to 36. 

### Other changes

None

### Optional section

- I didn't updated the `targetSdk` for `MobileSdk` and `MobileSdkRs`, not sure about the possible consequences for dependent apps. cc @sbihel 
- We need to release a new Showcase version before 31 Aug.

## Tested

- [X] Android app compiled successfully

### DIFF
--- a/android/Showcase/build.gradle.kts
+++ b/android/Showcase/build.gradle.kts
@@ -35,12 +35,12 @@ licensee {
 
 android {
     namespace = "com.spruceid.mobilesdkexample"
-    compileSdk = 35
+    compileSdk = 36
 
     defaultConfig {
         applicationId = "com.spruceid.mobilesdkexample"
         minSdk = 26
-        targetSdk = 35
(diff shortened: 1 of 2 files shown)


sbihel (approved):
I think it's good to update the SDK as well for consistency and to nudge dependent apps before they eventually get flagged by the Play Store, but no strong opinion

Juliano1612 (approved):
LGTM! I was checking and the Play Store requirement only applies to apps, not libraries. The only constraint is that a dependent app's `compileSdk` must be `≥` the library's, so libraries on 35 + apps on 36 is the safe direction. Bumping the SDKs to 36 wouldn't help with Play Store at all, but would force all dependent apps to upgrade to SDK 36. So I'd rather do that separately as a communicated change in a future release

PR #352, opened 2026-08-10
Acc 1781 implement vcalm flow in sprucekit showcase

aashishah-eng:
## Description
- Added VCALM protocol flows to SpruceKit Showcase for credential issuance and verification
    - This was built on the Rust protocol implementation in this repository, which remains unchanged  
- The UI for this implementation is minimal and can be adjusted to format information about the credential/exchange step
    - Used existing bottom sheets (Ex: ErrorView) and styling/colors

_Current Behaviour_
The current implementation deciphers the QR code scanned and decides whether to use the OID4VP/CI protocol or HTTP. It errors when taking in an interaction payload

_Updated Behaviour_
- Added support for interaction payloads - when this is scanned, protocol discovery is run, from which a bottom sheet shows the user all protocols supported to complete the interaction.
- If the user selects OID4VP/CI —> continue to existing flow for this protocol
- If the user selects vcapi —> the exchange for this protocol is kicked off 
    - POST to the exchange url, then handle presentation requests/offers/redirects appropriately
        - Offer: show offered credential to user, let them accept/decline. Once accepted, the credentials are added to the wallet (handed off to AddToWalletView, similar to other protocols) 
        - Request: error if no credentials meet the request, if there’s only one candidate auto-select the credential, otherwise allow the user to select which credential they’d like to submit here 
            - The user is told if all fields are necessary, else they see which fields are being showcased. They can also deselect optional fields if they wish (optional fields default to selected)
    - Complete the exchange loop until all outstanding requests are handled (empty body, which is surfaced to Swift/Kotlin as StepResult.Complete

### Other changes
- Added a bottom sheet in the case of a domain mismatch error (verifier’s domain doesn’t match exchange channel’s domain) —> this happens frequently on VCPlayground, so this sheet will help us override 
- A portion of AddToWalletView’s acceptCurrent function was extracted to a utility function, since that same behavior occurs in the VCALM flow
    - We can’t just use AddToWalletView immediately because once adding the credentials to the wallet, it returns home (meanwhile, the VCALM exchange loop may not be completed)

### Optional section

* Reviewers, please feel free to request updates to UI - as mentioned the UI is kept minimal for now 
* PR > 500 lines as it adds two views to handle the interaction, and then handle the VCALM protocol selection on both Kotlin and Swift implementations

## Tested
* Tested issuance of various credentials on VC Playground
* Tested verification of the credentials in permutations of the following scenarios: A) One credential matches vs Multiple credentials match and user selects the one to present B) Certain fields will be shared vs All fields will be shared 
* **This was tested by pointing to a specific branch of the `ssi` crate, as without it an AmbiguousProofs error was thrown upon credential issuance** 
The following was added to the end of Cargo.toml:
`[patch.crates-io]
ssi = { git = "https://github.com/spruceid/ssi", branch = "fix/multiproof-select-0.14" }
ssi-jwk = { git = "https://github.com/spruceid/ssi", branch = "fix/multiproof-select-0.14" }
ssi-jws = { git = "https://github.com/spruceid/ssi", branch = "fix/multiproof-select-0.14" }
`

**Issuance and verification (one credential matches, all fields required)**
https://github.com/user-attachments/assets/069915ca-6777-4773-9f05-97bdaceff493

**Credential selector if multiple match**
image

**Field selector on verification flow**
image


SeaSkyThe (changes requested):
EDIT: you can set the images sizes by deleting the html height property and setting width to 250, the engine will render it by keeping the aspect ratio.

Tested only on iOS for now. Functionality wise, everything seems to be working. I'll focus this first review on UI tweaks we can make. If something I said is not possible due to protocol limitations, please let me know.

1. Found a bug where whenever we close the protocol selection bottom sheet, the user gets stuck in the "Fetching Protocols" screen.
     - Also, for this screen, you could probably reuse the `LoadingView`component.

2. Another "bug" in the Verifier Domain Mismatch screen. Whenever I click cancel in the modal, nothing happens, the user is forced to continue the flow. We should probably close the how flow and show an error modal saying something like "Presentation flow canceled". And if possible, can you test other "user cancelled" flows to make sure we didn't forget anything?

3. Can we try to make the VC-API flow similar to OID4VP/VCI style-wise?
     - Test OID4VCI/VP with [Animo Playground Issuer (EU PID sd-jwt)](https://playground.animo.id/?authorization=none&credentialType=1&deferBy=none&dpop=false&format=dc%2Bsd-jwt&issuerId=188e2459-6da8-4431-9062-2fcdac274f41&keyAttestation=false&tab=issue&walletAttestation=false). And [Animo Playground Verifier (EU PID sd-jwt)](https://playground.animo.id/?presentationDefinitionId=8caaebcc-d48c-471b-86b0-a534e15c4774__1&requestScheme=openid4vp%3A%2F%2F&requestSignerType=x5c&responseMode=direct_post.jwt&tab=verify&transactionAuthorizationType=none).
     - Issuance VC-API
          - I think we can get rid of the "Credential offer" screen (see below), it feels like a duplicate of the screen that comes right after it (the "Review Info" one). Is that possible?
             Image

     - Presentation VC-API
          - When we have multiple credentials, the credential selection screen could be similar to its OID4VP equivalent (padding, title, checkbox, etc). My impression is that we are mixing the Credential Selection (when having multiple credentials of the type requested) with the Selective Disclosure screen?

          <table>
            <tr>
              <th>VC-API</th>
              <th>OID4VP</th>
            </tr>
            <tr>
              <td>Image</td>
              <td>Image</td>
             </tr>
          </table>

          - Lets get rid of the success screen, I think we can go simpler and go back to home and show a toast with a "Shared successfully" message on success. Just like the OID4VP flow.

     - Modal style mismatch
          - The Verifier Domain Mismatch and Protocol Selection modals are different in style. Let’s use the Protocol Selection one as the base for both (create a shared component if we don’t already have one), but make the Title and Subtitle different styles. The title should be bigger and bold, while the subtitle should be smaller, have a clearer color, and have some padding from the title.

aashishah-eng:
@SeaSkyThe Addressed the feedback, Here are the screen recordings of issuance/verification on both android and iOS:

https://github.com/user-attachments/assets/846be17d-dc51-4bdd-9651-4b2bb43702d3


 
Android:


https://github.com/user-attachments/assets/110d21be-20fa-4960-8c7d-67676ff26fcd


Also tested all flows where cancellation occurs, they either show an error view as we discussed, or a toast and redirect the user to the home page

SeaSkyThe (commented):
Hey. It looks good to me. I'll ask @Juliano1612 for a last check and after that I think we are good to merge.

android/Showcase/src/main/java/com/spruceid/mobilesdkexample/wallet/ProtocolSelectorBottomSheet.kt
```diff
@@ -0,0 +1,206 @@
...
+                    uri.startsWith(MDOC_OID4VP_SCHEME) && !credentialPackId.isNullOrEmpty() ->
+                        Screen.HandleMdocOID4VPWithCredentialPack.route.replace(
+                            "{credential_pack_id}",
+                            credentialPackId
+                        )
+
+                    uri.startsWith(MDOC_OID4VP_SCHEME) ->
+                        Screen.HandleMdocOID4VP.route
+
+                    else -> throw IllegalArgumentException("Invalid OID4VP scheme")
```
Juliano1612: this `throw` runs inside a Compose `onClick`, and an OID4VP entry whose `URI` doesn't start with `openid4vp://` or `mdoc-openid4vp://` crashes the app. We should show an error state instead

android/Showcase/src/main/java/com/spruceid/mobilesdkexample/wallet/DispatchQRView.kt
```diff
@@ -70,13 +73,34 @@ fun DispatchQRView(
...
     }
 
+    fun isInteraction(url: String): Boolean {
+        val stripped = if (url.startsWith("interaction:", ignoreCase = true)) {
+            url.substring("interaction:".length)
+        } else {
+            url
+        }
+
+        val query: String = URI(stripped).query ?: ""
```
Juliano1612: `URI(stripped)` throws on any payload that isn't a parseable URI, and this runs first in the `when` chain. Wrap this in a `try/catch` that returns false, like the iOS `URLComponents` version already does

android/Showcase/src/main/java/com/spruceid/mobilesdkexample/wallet/HandleVCALMView.kt
```diff
@@ -0,0 +1,1005 @@
...
+                    onSuccess = {
+                        pendingWalletCredentials = null
+                        when (val result = offerAcceptResult) {
+                            null -> coroutineScope.launch { declineOffer() }
+                            is StepResult.Complete ->
+                                navController.navigate(Screen.HomeScreen.route) { popUpTo(0) }
+
+                            is StepResult.Problem -> {}
+                            // Chained — the exchange isn't done yet, continue to whatever's next.
+                            else -> coroutineScope.launch { handleStep(result) }
```
Juliano1612, line 465: This `launch { handleStep(result) }` is the only call site with no `try/catch` around it. Can you wrap `onRequest`'s body in `try/catch` like the Swift version does?

android/Showcase/src/main/java/com/spruceid/mobilesdkexample/wallet/HandleVCALMView.kt
```diff
@@ -0,0 +1,1005 @@
...
+                        acceptRawCredentialIntoWallet(
+                            raw,
+                            credentialPacksViewModel,
+                            walletActivityLogsViewModel,
+                        )
+                    },
+                    onSuccess = {
+                        pendingWalletCredentials = null
+                        when (val result = offerAcceptResult) {
+                            null -> coroutineScope.launch { declineOffer() }
```
Juliano1612: when `acceptOffer()` fails, `offerAcceptResult` stays null, so this branch POSTs a reject on top of the failed accept and navigates home with an "Offer declined" success toast. Bail out here when `offerAcceptError` is set. This also happens in the iOS flow

android/Showcase/src/main/java/com/spruceid/mobilesdkexample/wallet/HandleVCALMView.kt
```diff
@@ -0,0 +1,1005 @@
...
+            // Treat this as exchange ended, and navigate to home screen
+            Toast.showSuccess("Offer declined")
+            navController.navigate(Screen.HomeScreen.route) { popUpTo(0) }
+        }
+    }
+
+    suspend fun startExchange() {
+        try {
+            val vdcCollection = VdcCollection(engine = credentialPacksViewModel.storageManager)
+            val signer = VCALMSigner("vcalm_holder_key")
```
Juliano1612: can we use the `DEFAULT_SIGNING_KEY_ID`here instead? The way it is now, credentials issued through VCALM get a different holder DID than everything else in the app, so presenting them through another flow (or vice versa) can fail holder-binding checks

android/Showcase/src/main/java/com/spruceid/mobilesdkexample/wallet/HandleVCALMView.kt
```diff
@@ -0,0 +1,1005 @@
...
+
+    override fun jwk(): String {
+        return jwk
+    }
+
+    override fun cryptosuite(): String {
+        // TODO: Add an uniffi enum type for crypto suites.
+        return "ecdsa-rdfc-2019"
+    }
+}
```
Juliano1612: This is basically the `Signer` in `HandleOID4VPView` (which already takes a keyId). Let's reuse the existing one instead of duplicating it.

Juliano1612: Same for iOS

android/Showcase/src/main/java/com/spruceid/mobilesdkexample/wallet/HandleVCALMView.kt
```diff
@@ -0,0 +1,1005 @@
...
+fun vcalmCredentialIssuedDate(
+    parsedCredential: ParsedCredential,
+    credentialClaims: Map<String, JSONObject>,
+): String? {
+    val claims = credentialClaims[parsedCredential.id()] ?: return null
+    val raw = claims.optString("validFrom").takeIf { it.isNotBlank() }
+        ?: claims.optString("issuanceDate").takeIf { it.isNotBlank() }
+        ?: return null
+    return raw.substringBefore("T")
+}
```
Juliano1612: looks like this is not being used on both platforms. Can you remove it or wire it in the right place?

android/Showcase/src/main/java/com/spruceid/mobilesdkexample/wallet/HandleVCALMView.kt
```diff
@@ -0,0 +1,1005 @@
...
+        if (url.startsWith(scheme)) {
+            val remainder = url.substring(scheme.length).replaceFirst(Regex("^/+"), "")
+            try {
+                val decoded = Uri.decode(remainder)
+                if (decoded.startsWith("http://") || decoded.startsWith("https://")) {
+                    url = decoded
+                }
+            } catch (_: Error) {
+                // Not a percent-encoded URL - leave as-is
+            }
```
Juliano1612: I believe you should use `Exception` here instead of `Error`. In Kotlin, Error is the class for fatal VM errors, not a common exception. So catch `Exception`, or drop the try/catch since `Uri.decode` is lenient anyway

android/Showcase/src/main/java/com/spruceid/mobilesdkexample/wallet/HandleVCALMView.kt
```diff
@@ -0,0 +1,1005 @@
...
+        val q = encoded.indexOf("?")
+        if (q >= 0) encoded = encoded.substring(0, q)
+        try {
+            val decoded = Uri.decode(encoded)
+            if (decoded.startsWith("http://") || decoded.startsWith("https://")) {
+                return decoded
+            }
+        } catch (_: Error) {
+            // Not an encoded URL - use existing input
+        }
```
Juliano1612: Same from 298 here

android/Showcase/src/main/java/com/spruceid/mobilesdkexample/wallet/HandleVCALMView.kt
```diff
@@ -0,0 +1,1005 @@
...
+                    listOf("name", "type", "validFrom", "issuanceDate")
+                )
+            }
+            holder!!.provideCredentials(credentials)
+
+            val result = holder!!.startExchange(unwrap(url), null)
+
+            handleStep(result)
+        } catch (e: Exception) {
+            errorTitle = "Error Adding Credential"
```
Juliano1612: Do we already know it's an issue flow at this moment? Otherwise something like "Error Starting Exchange" should fit better

Juliano1612: same on iOS

ios/Showcase/Targets/AppUIKit/Sources/wallet/ProtocolSelectorBottomSheet.swift
```diff
@@ -0,0 +1,137 @@
...
+                )
+            } else if uri.hasPrefix(MDOC_OID4VP_SCHEME) {
+                path.append(
+                    HandleMdocOID4VP(
+                        url: uri,
+                        credentialPackId: credentialPackId
+                    )
+                )
+            }
+        case "vcapi", _:
```
Juliano1612: this `, _` is silently routing any future protocol to the VCALM handler. Let's keep just `case "vcapi"` and handle default explicitly

android/Showcase/src/main/java/com/spruceid/mobilesdkexample/wallet/HandleVCALMView.kt
```diff
@@ -0,0 +1,1005 @@
...
+
+        Column(
+            modifier = Modifier
+                .fillMaxSize()
+                .verticalScroll(rememberScrollState())
+                .weight(weight = 1f, fill = false)
+        ) {
+            if (currentRequirement.fields.isEmpty()) {
+                // No specific fields requested, show all claims from the credential
+                val allClaims = currentCredential?.let { credentialClaims[it.id()] } ?: JSONObject()
```
Juliano1612, line 788: `credentialClaims` only ever holds name/type/validFrom/issuanceDate, because `startExchange`is filtering to get only these claims. So this "show all claims" branch under-reports what actually gets disclosed. Can you change to fetch the full claim set for the consent screen?

Juliano1612: same on iOS

android/Showcase/src/main/java/com/spruceid/mobilesdkexample/wallet/DispatchQRView.kt
```diff
@@ -88,6 +112,18 @@ fun DispatchQRView(
                 // Check if payload type is supported
                 if (qrType != null && supportedTypes.contains(qrType)) {
                     when (qrType) {
+                        SupportedQRTypes.INTERACTION -> {
+                            val route =
+                                Screen.HandleInteraction.route.replace("{url}", encodedUrl).replace(
+                                    "{credential_pack_id}",
+                                    credentialPackId ?: "null"
```
Juliano1612: this `credentialPackId ?: "null"` puts the literal string "null" in the route, `ProtocolSelectorBottomSheet` then passes `isNullOrEmpty()` check and routes to `HandleOID4VPWithCredentialPack` with pack id "null". We should map it back to null when reading the nav argument, or split into two routes like the existing OID4VP ones

Juliano1612 (changes requested):
Really solid work! The exchange loop handling is clean. I'm requesting changes mainly on the consent screens, since those need to be truthful about what gets shared. Everything else in the inline comments is smaller.

Juliano1612 (approved):
LGTM!

SeaSkyThe (approved):

PR #354, opened 2026-08-17
Acc 1780 extract vcalm out into its own repo

aashishah-eng:
## Description

- Currently, sprucekit-mobile relies on the `rust/src/vcalm` files to access this protocol
- By extracting these into a [standalone crate](https://github.com/spruceid/vcalm-rs), other crates that have this protocol implemented can depend on the `vcalm-rs` crate directly, rather than having to use the entirety of `mobile-sdk-rs` as a dependency (which may have significantly more unused functionality in this application) 

_Implemented vcalm-rs similarly to [oid4vci](https://github.com/spruceid/oid4vci-rs) and [openid4vp](https://github.com/spruceid/openid4vp), to maintain consistency_ 
- Existing VCALM implementation used uniffi types (whereas data types never crossed the FFI in OID4VCI/VP)
- These types were converted into objects, and accessors were added to the Swift/Kotlin layers 
- ex: () was added to access fields of structs in HandleVCALMView.kt

_Moving `discover_protocols` to the VCALM crate_
- Vcalm depended on this method, by leaving it in sprucekit-mobile there would’ve been dependencies back and forth 
    - The refactor we did [earlier](https://github.com/spruceid/sprucekit-mobile/pull/349) brought functions shared by `vcalm` and `discover_protocols` to the `discover_protocols`' file
    - Implications of doing this
        - on the Swift and Kotlin side, the bindings in MobileSdkRs cover `discover_protocols`, so no change there
        - Rust, if a host crate depends on mobile-sdk-rs: `use mobile_sdk_rs::discover_protocols::{discover_protocols, DiscoveryError};`
        - Rust, if a host depends directly on vcalm-rs: `use vcalm_rs::discover_protocols::discover_protocols;`
    - If we still want to keep `discover_protocols` in sprucekit, we can move the entry points back

### Other changes
N/A

### Optional section
N/A

## Tested
- Completed pre-push checklist (`cargo build`, `cargo test`, etc.)
- Tested issuance and verification flows on iOS and Android
- **Like the initial VCALM implementation in sprucekit-mobile, a patch needs to be added for successful testing:**
This was tested by pointing to a specific branch of the ssi crate, as without it an AmbiguousProofs error was thrown upon credential issuance
The following was added to the end of Cargo.toml:
`[patch.crates-io] ssi = { git = "https://github.com/spruceid/ssi", branch = "fix/multiproof-select-0.14" } ssi-jwk = { git = "https://github.com/spruceid/ssi", branch = "fix/multiproof-select-0.14" } ssi-jws = { git = "https://github.com/spruceid/ssi", branch = "fix/multiproof-select-0.14" } `


rust/src/vcalm/wire.rs
```diff
@@ -0,0 +1,462 @@
...
+use vcalm_rs::exchange as vc;
+use vcalm_rs::holder as vh;
+
+/// Render an optional JSON value as its compact text form.
+fn json_text(value: Option<serde_json::Value>) -> Option<String> {
+    value.map(|v| v.to_string())
+}
+
+/// A verifiable-presentation-request (§3.4).
+#[derive(Debug, uniffi::Object)]
```
Juliano1612: These wrappers could be `uniffi::Record` mirrors instead of Objects with accessors. One deep conversion at the boundary, and the `0.20.x` Kotlin/Swift API stays intact (no `()` churn across every consumer).

You can keep this and bump to v0.21.x or revert

rust/src/vcalm/mod.rs
```diff
@@ -1,11 +1,176 @@
...
+    #[uniffi::constructor]
+    pub async fn new_session(
+        vdc_collection: Arc<VdcCollection>,
+        trusted_dids: Vec<String>,
+        signer: Box<dyn PresentationSigner>,
+        context_map: Option<HashMap<String, String>>,
+        keystore: Option<Arc<dyn KeyStore>>,
+    ) -> Result<Arc<Self>, VcalmError> {
+        // `keystore` is accepted and ignored (dropping the parameter would change the generated bindings)
+        let _ = &keystore;
```
Juliano1612: The "dropping the parameter would change the generated bindings" rationale doesn't hold up. This PR already regenerates the bindings and changes the VCALM API surface (Records > Objects). If we're taking an API break anyway, this is the moment to drop the dead parameter

rust/src/credential/format/json_vc.rs
Juliano1612: `derive_sd_vp_credential` was moved to `vcalm-rs` but its error type stayed behind

rust/src/vcalm_adapters.rs
```diff
@@ -0,0 +1,207 @@
...
+        let body = parsed
+            .as_json_vc()
+            .map(|json_vc| json_vc.raw.clone())
+            .unwrap_or(Value::Null);
+
+        Ok(Some(StoredCredential {
+            id,
+            body,
+            host: parsed,
+        }))
```
Juliano1612: This body projection (`as_json_vc()` -> `raw / Null`) is the same logic as `to_stored` in `vcalm/mod.rs`. Worth extracting a single helper so the "Null means not matchable" rule lives in one place with one caveat: here the storage id is authoritative, while `to_stored` trusts `credential.id()`. If those can ever diverge, keep the explicit `id`

rust/src/vcalm/mod.rs
```diff
@@ -1,11 +1,176 @@
...
+
+#[derive(uniffi::Record)]
+pub struct VcalmMatchedCredential {
+    pub credential: Arc<ParsedCredential>,
+    pub selective_disclosure: bool,
+}
+
+pub fn stable_local_id(entry: &Value) -> Uuid {
+    vcalm_rs::issuance::stable_local_id(entry)
+}
```
Juliano1612: This wrapper isn't `#[uniffi::export]`ed and has no callers in the crate

rust/src/vcalm/wire.rs
```diff
@@ -0,0 +1,462 @@
...
+    SdDeriveFailed(String),
+    #[error("unsupported credential format: {0}")]
+    UnsupportedCredentialFormat(String),
+}
+
+impl From<vcalm_rs::error::VcalmError> for VcalmError {
+    fn from(e: vcalm_rs::error::VcalmError) -> Self {
+        use vcalm_rs::error::VcalmError as E;
+        match e {
+            E::UnexpectedUniFFICallbackError(v) => Self::UnexpectedUniFFICallbackError(v),
```
Juliano1612: `vcalm-rs` is uniffi-free, so its `UnexpectedUniFFICallbackError` variant can never be constructed upstream

Juliano1612 (changes requested):
verified discover_protocols parity and the adapter layer looks good. Two things before merging:

- Records > Objects breaks the `0.20.x` Kotlin/Swift API and adds an FFI hop per field read 
- Smaller cleanups were flagged

rust/src/vcalm_adapters.rs
aashishah-eng: Moved into a helper function but kept id explicit, in case the parsed credential is built without threading the stored id, which would cause a silent failure

rust/src/vcalm/wire.rs
aashishah-eng: Reverted to `uniffi::Record`, but I believe it will still need to be bumped to v0.21.0 due to the dead param drop listed below, along with VcalmError having different types now.

rust/Cargo.toml
```diff
@@ -101,3 +102,8 @@ wiremock = "0.6.3"
 
 [build-dependencies]
 uniffi = { workspace = true, features = ["build"] }
+
+[patch.crates-io]
```
Juliano1612, line 108:
```suggestion

# Temporary: required for VCALM issuance (AmbiguousProofs on multi-proof VCs).
# Drop once the fix ships in a crates.io release.
[patch.crates-io] 
```

Juliano1612 (approved):
Looks good to me! All feedback addressed. Just added one suggestion but it's 100% cosmetic

PR #356, opened 2026-08-20
Fix wallet crash on credentials with http image links

leoliu-spruce:
[SKIT-852
](https://linear.app/spruceid/issue/SKIT-852/fix-android-wallet-crash-with-http-image-links)

## Description

`CredentialImage` only treated `https://` values as remote links, so an `http://` image fell through to the branch that decodes inline base64. `Base64.decode` skips characters outside the base64 alphabet instead of rejecting them, so `http://example.com/logo.png` lost its `:` and `.` and decoded cleanly into 18 bytes of garbage. BitmapFactory returned null for those bytes, and BitmapImage dereferenced the result through a non-null `Bitmap` return type, so opening the wallet crashed with `NullPointerException: decodeByteArray(...) must not be null`.

## Tested

Added a `CredentialImageTest.kt` to test all 6 branches

### DIFF
--- a/android/Showcase/src/androidTest/java/com/spruceid/mobilesdkexample/credentials/CredentialImageTest.kt
+++ b/android/Showcase/src/androidTest/java/com/spruceid/mobilesdkexample/credentials/CredentialImageTest.kt
@@ -1,0 +1,81 @@
+package com.spruceid.mobilesdkexample.credentials
+
+import androidx.compose.foundation.layout.Box
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithContentDescription
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+/**
+ * Needs a device: `android.util.Base64` and `BitmapFactory` are unavailable to JVM tests.
+ */
+@RunWith(AndroidJUnit4::class)
+class CredentialImageTest {
+    @get:Rule
+    val composeTestRule = createComposeRule()
+
+    /** A 1x1 transparent PNG. */
+    private val pngBase64 =
+        "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNkYPhfDwAChwGA60e6kgAAAABJRU5ErkJggg=="
+
+    /** Renders inside a tagged container so an absent image can be told from an absent hierarchy. */
+    private fun setContentInContainer(image: String, alt: String) {
+        composeTestRule.setContent {
+            Box(Modifier.testTag("container")) {
+                CredentialImage(image, alt)
+            }
+        }
+        composeTestRule.onNodeWithTag("container").assertExists()
+    }
+
+    @Test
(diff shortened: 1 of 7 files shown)


ios/Showcase/Targets/AppUIKit/Sources/wallet/WalletUtils.swift
```diff
@@ -53,6 +53,12 @@ extension String {
     func replaceEscaping() -> String {
         return self.replacingOccurrences(of: "\\/", with: "/")
     }
+
+    /// Whether the value links to a remotely hosted image rather than holding an inline base64 one.
+    var isRemoteImageUrl: Bool {
+        let value = self.lowercased()
```
Juliano1612: `lowercased()` copies the whole string, and inline base64 payloads can run to hundreds of KB

```suggestion
        let value = self.prefix(8).lowercased()
```

android/Showcase/src/androidTest/java/com/spruceid/mobilesdkexample/credentials/CredentialImageTest.kt
```diff
@@ -0,0 +1,81 @@
+package com.spruceid.mobilesdkexample.credentials
```
Juliano1612, line 1: Heads up: CI only runs `:showcase:test`, so this suite doesn't run automatically. Fine to land as is, but worth adding a step so these actually gate regressions.

android/Showcase/src/main/java/com/spruceid/mobilesdkexample/credentials/CredentialImage.kt
```diff
@@ -11,12 +10,13 @@ import androidx.compose.ui.layout.ContentScale
...
 import coil3.compose.AsyncImage
 import com.spruceid.mobilesdkexample.utils.BitmapImage
+import com.spruceid.mobilesdkexample.utils.isRemoteImageUrl
 import com.spruceid.mobilesdkexample.utils.jsonArrayToByteArray
 import org.json.JSONArray
 
 @Composable
 fun CredentialImage(image: String, alt: String) {
-    if (image.startsWith("https://")) {
+    if (image.isRemoteImageUrl()) {
```
Juliano1612, line 19: This fixes the crash, but the `http` image still won't load, even in debug builds because `cleartext` is blocked by `targetSdk`. Since `http://` is only meant for testing against a local backend, we can create a debug-only manifest to unlock that without touching release:

`Showcase/src/debug/AndroidManifest.xml`

```
<?xml version="1.0" encoding="utf-8"?>
<manifest xmlns:android="http://schemas.android.com/apk/res/android">
    <application android:usesCleartextTraffic="true" />
</manifest>
```

Juliano1612 (changes requested):

android/Showcase/src/androidTest/java/com/spruceid/mobilesdkexample/credentials/CredentialImageTest.kt
leoliu-spruce: For the suite not running in the CI, there are two options and let me know which one sounds reasonable:
1. Leaving it local as it is now, so no additional testing setup but also no CI regression gates
2. Use an advisory emulator job which is the same pattern as safe-mobile's android-emulator: runs on every PR, kept out of required checks, red result visible but never blocks. This will be slow as it involves building the app, installing and running it in an emulator

If we decide to set up an emulator, I can make a separate issue for that

Juliano1612: Let's keep it local. We can add it to the CI in the future as we add more tests.

Juliano1612 (approved):

PR #358, opened 2026-08-21
Add parameter to pass selected fields to submit_presentation

aashishah-eng:
## Description
- See associated [PR on `vcalm-rs`](https://github.com/spruceid/vcalm-rs/pull/2)
- Implemented UI such as sprucekit-mobile's Showcase lets users uncheck optional fields, but this isn't ever passed to the protocol (`VcalmHolder::submit_presentation`) to be referenced during SD derive
- This PR propogates the state tracking selected fields to the HandleVCALMView level, so it can be passed into `VcalmHolder::submit_presentation`

## Tested
- Only selected fields from UI are being passed to `submit_presentation`
- Unit test in the rust layer ensures that deselected fields are not disclosed

### DIFF
--- a/android/Showcase/src/main/java/com/spruceid/mobilesdkexample/wallet/HandleVCALMView.kt
+++ b/android/Showcase/src/main/java/com/spruceid/mobilesdkexample/wallet/HandleVCALMView.kt
@@ -193,6 +193,8 @@ fun HandleVCALMView(
     var credentialClaims by remember { mutableStateOf(mapOf<String, JSONObject>()) }
 
     var requirements by remember { mutableStateOf<List<VcalmRequirement>?>(null) }
+    // Field paths the user consents to disclose, keyed by queryIndex
+    var fieldSelections by remember { mutableStateOf<Map<UInt, Set<String>>>(emptyMap()) }
     var picks by remember { mutableStateOf<Map<UInt, ParsedCredential>>(emptyMap()) }
     var redirectUrl by remember { mutableStateOf<String?>(null) }
     var pendingWalletCredentials by remember { mutableStateOf<List<String>?>(null) }
@@ -200,6 +202,7 @@ fun HandleVCALMView(
     var offerAcceptError by remember { mutableStateOf(false) }
     var domainMismatch by remember { mutableStateOf<VcalmException.DomainChannelMismatch?>(null) }
     var pendingSelection by remember { mutableStateOf<List<ParsedCredential>>(emptyList()) }
+    var pendingFieldPaths by remember { mutableStateOf<Map<UInt, List<String>>>(emptyMap()) }
 
     fun unwrap(originalUrl: String): String {
         var url = originalUrl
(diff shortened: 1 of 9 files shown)


android/Showcase/src/main/java/com/spruceid/mobilesdkexample/wallet/HandleVCALMView.kt
```diff
@@ -805,20 +824,20 @@ fun VcalmFieldsSelector(
                     }
                 }
             } else {
-                var selectedFields by remember(currentRequirement.queryIndex) {
-                    mutableStateOf(currentRequirement.fields.map { it.path }.toSet())
-                }
+                // Field toggles propogate to HandleVCALMView
```
Juliano1612:
```suggestion
                // Field toggles propagate to HandleVCALMView
```

android/Showcase/src/main/java/com/spruceid/mobilesdkexample/wallet/HandleVCALMView.kt
```diff
@@ -341,8 +350,12 @@ fun HandleVCALMView(
     }
 
     suspend fun submitPicks() {
-        val selected = requirements.orEmpty().mapNotNull { picks[it.queryIndex] }
-        trySubmitPresentation(selected, false)
+        val reqs = requirements.orEmpty()
+        val selected = reqs.mapNotNull { picks[it.queryIndex] }
+        val selectedFields = reqs.associate {
+            it.queryIndex to fieldSelections[it.queryIndex].orEmpty().toList()
+        }
```
Juliano1612: Small but important one. We send an entry for every requirement, so a type-only query with no checkboxes goes out as an explicit empty list. Today that happens to work because the rust side rescues type-only queries via `present_named_subject`, but on your VCALM PR we're asking for `missing key = no narrowing` and `explicit empty list = deselect all`. Under that contract this line would be claiming "the user deselected everything" when there was nothing to select in the first place. Let's only send entries for requirements that actually had checkboxes, a `reqs.filter { it.fields.isNotEmpty() }` before the associate does it. The iOS `submitPicks` needs the same tweak

rust/src/vcalm/mod.rs
```diff
@@ -108,13 +108,15 @@ impl VcalmHolder {
     pub async fn submit_presentation(
```
Juliano1612, line 130: This wrapper is where the generated Swift and Kotlin docs come from and it still reads like the old two-param version. Once the semantics settle on your VCALM PR, let's mirror the `selected_fields` doc from `holder.rs` here: keyed by query index, missing key vs explicit empty list, paths must match `requested_fields()` exactly, and narrowing only applies to credentials with an `ecdsa-sd-2023` base proof.

Juliano1612 (changes requested):

flutter/android/src/main/kotlin/com/spruceid/sprucekit_mobile/VcalmAdapter.kt
```diff
@@ -267,7 +267,6 @@ internal class VcalmAdapter(
                 // Suite is server-driven — no suite parameter.
                 val creds = selected.mapNotNull { keyMap[it] }
                 Log.d(TAG, "submitPresentation: resolved ${creds.size}/${selected.size} handle(s)")
-                val step = h.submitPresentation(creds, allowDomainMismatch)
```
Juliano1612, line 270: Any reason to remove the submission?

Juliano1612 (approved):
Approved! But I believe you need to restore the presentation submission as I commented on your `VcalmAdapter.kt`.

PR #360, opened 2026-08-26
Allow user to deny consent to share credentials during presentation

patmalt:
## Description

A user can decide to deny consent while presenting their credentials. The verifier app needs to be made aware of this choice by the user so it can clean up local state and display an error message. Without responding to the denial, the verifier app will be left on its own and require a time out or user intervention.

image

The `access_denied` error code can specify that “[The End-User did not give consent to share the requested Credentials with the Verifier](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-8.5-9.2)”.

Exposes `Holder` `denyPermission` API that will POST `/response` with `error=access_denied` and `state={state}`, which informs verifier that the user cancelled.

### Other changes

Multiple `Holder` rust implementations when wallet uses the `AUTO` OID4VP compatibility. 

## Tested

- Ran same device and cross device credential presentation from iOS and Android consumer of `SpruceIDMobileSdk`.
- Build and run iOS, Android, and Flutter showcase apps.


Ryanmtate (approved):
Nice work! 👍

PR #368, opened 2026-09-17
Use native HttpClient for Rust layer and move reqwest as fallback

w4ll3:
## Description

Replaces CIO in favor of HttpClient for Android, even though CIO is optmized for pure Kotlin environments OkHttp is much more mature and battle tested, for Android makes sense for use to use it. The main reason this change is coming right now is because CIO does not honor the system configured proxy, which makes your app fail if you have that on.

Moves `reqwest` as a fallback for the native `HttpClient`.

### Other changes

This changes came mostly from the need of auditors to use proxy, after understanding the why decided to make the structural changes not only to add that support but to also improve the library

## Tested

Tested on Android and iOS, both worked fine. Would be great to have coverage on other platforms that are being targeted.


flutter/android/build.gradle
```diff
@@ -19,7 +19,7 @@ allprojects {
     repositories {
         // Uncomment mavenLocal() when testing against a locally-built SDK artifact.
         // See flutter/CLAUDE.md → "Adding a new Rust function to the Android adapter".
-        // mavenLocal()
+        mavenLocal()
```
jessmelanson, line 22: Guessing this should remain commented out?

jessmelanson (approved):
Looks good, don't see any impacts for client app

Ryanmtate (approved):

